PR #70, opened 2015-01-20
Configure receive timeout to QC and DC clients.

chitang:


soda-fountain-lib/src/main/resources/reference.conf
```diff
@@ -34,11 +34,13 @@ com.socrata.soda-fountain = {
     service-name = "data-coordinator"
     instance = "alpha"
     connect-timeout = 1s
+    receive-timeout = 1day
```
sokrahta: probably not important, but the first intent of no timeout could be closer achieved using ```24day``` which is near int max for milliseconds.

sokrahta:
:shipit:

PR #80, opened 2015-03-07
Add support so we can reference the database password in a seperate file to enable clortho key management support.

znep:


velvia:
Looks fine to me (the more interesting bits are under the covers).  Going to merge.

PR #82, opened 2015-03-12
Computation strategy fix (CORE-4551)

urmilan:
Fixing a bug in Soda Fountain where the computation_strategy returned from the /dataset/{resourceName} endpoint returns the source column's internal ID of instead of its SoQL field name.
cc: @cmlaidlaw


urmilan:
Another OOM...

urmilan:
If I do that I'll have to manually write out the codecs for all the dependent case classes below, instead of being able to use `AutomaticJsonCodecBuilder`. I'd rather not write so much boilerplate.

PR #83, opened 2015-03-13
BUG: Cannot create column with computation strategy.

chitang:
Stan found a bug where we cannot create column with computation strategy.


znep:
I'm a little worried about the lack of test coverage on this... but given other things going on right now I think it is reasonable to merge anyway.

PR #87, opened 2015-03-28
Spandex2

sokrahta:
push suggest requests from `core` through `soda-fountain` to `spandex`

[Review on Reviewable](https://reviewable.io/reviews/socrata/soda-fountain/87)

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
@@ -164,6 +164,9 @@ class DatasetDAOImpl(dc: DataCoordinatorClient, store: NameAndSchemaStore, colum
         NotFound(dataset)
     }
 
+  def getCurrentCopyNum(dataset: ResourceName): Option[Long] =
+    store.lookupCopyNumber(dataset, None)
+
   def getDataset(dataset: ResourceName, stage: Option[Stage]): Result =
     store.lookupDataset(dataset, stage) match {
       case Some(datasetRecord) => Found(datasetRecord)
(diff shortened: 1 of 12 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
```diff
@@ -164,6 +164,9 @@ class DatasetDAOImpl(dc: DataCoordinatorClient, store: NameAndSchemaStore, colum
         NotFound(dataset)
     }
 
+  def getCurrentCopyNum(dataset: ResourceName): Option[Long] =
+    store.lookupCopyNumber(dataset, None) // stage=None implies latest copy
```
znep: We should be using the latest published copy if there is one, otherwise the unpublished copy.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Suggest.scala
```diff
@@ -0,0 +1,70 @@
...
+  lazy val spandexAddress = s"${config.host}:${config.port}"
+
+  def datasetId(resourceName: ResourceName): Option[String] = {
+    datasetDao.getDataset(resourceName, None) match {
+      case DatasetDAO.Found(d) => Some(d.systemId.underlying)
+      case DatasetDAO.NotFound(d) => None
+      case x: DatasetDAO.Result =>
+        val msg = s"dataset not found $resourceName ${x.getClass.getName}"
+        log.error(msg)
+        throw new Exception(msg)
```
znep: Shouldn't this return a 404?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
sokrahta: fixed in #91

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Suggest.scala
sokrahta: If the dataset is `NotFound` then `None` comprehends to a 404 in the `get` call.

PR #89, opened 2015-03-28
es mimetype is not json

chitang:
report 404 if something isn't found.


znep:
Should we just have spandex return the proper mime type?  Or am I confused.

sokrahta:
spandex is returning the proper mime type after https://github.com/socrata/spandex/pull/8
lgtm :shipit:

PR #91, opened 2015-03-31
Spandex

sokrahta:
+ current copy is now latest published copy
+ fuzziness/size parameters passed through to spandex

[Review on Reviewable](https://reviewable.io/reviews/socrata/soda-fountain/91)

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
@@ -165,7 +165,7 @@ class DatasetDAOImpl(dc: DataCoordinatorClient, store: NameAndSchemaStore, colum
     }
 
   def getCurrentCopyNum(dataset: ResourceName): Option[Long] =
-    store.lookupCopyNumber(dataset, None) // stage=None implies latest copy
+    store.lookupCopyNumber(dataset, Some(Published))
 
   def getDataset(dataset: ResourceName, stage: Option[Stage]): Result =
     store.lookupDataset(dataset, stage) match {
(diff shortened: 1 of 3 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
```diff
@@ -165,7 +165,7 @@ class DatasetDAOImpl(dc: DataCoordinatorClient, store: NameAndSchemaStore, colum
     }
 
   def getCurrentCopyNum(dataset: ResourceName): Option[Long] =
-    store.lookupCopyNumber(dataset, None) // stage=None implies latest copy
+    store.lookupCopyNumber(dataset, Some(Published))
```
znep: You also need to account for the case where there are no published copies, ie. the only copy is Unpublished.

urmilan:
Hooray for a DAO test! :+1:

soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
sokrahta: If I grok this correctly: when there are no published copies the lookup will return `None`. I believe that is the correct behavior, we don't want to expose an unpublished copy just because there are no published copies, yes?

znep: Yes, we do want to expose an unpublished copy if there is no published copy.

The code that does it for the pg secondary is at
https://github.com/socrata/soql-postgres-adapter/blob/master/soql-server-pg/src/main/scala/com/socrata/pg/server/QueryServer.scala#L371-L383
for reference.  Not that I'm saying that code is great code or you should
just copy it, but it shows the behavior we currently have.

sokrahta: okay thanks! fixed in bdb8e0c

PR #92, opened 2015-03-31
Reduce queue length to keep request latency lower

velvia:
At 500, we should be able to keep /version running.  Might have to tune it some more, but better than the previous value.

### DIFF
--- a/soda-fountain-lib/src/main/resources/reference.conf
+++ b/soda-fountain-lib/src/main/resources/reference.conf
@@ -92,8 +92,8 @@ com.socrata.soda-fountain = {
     min-threads = 10
     max-threads = 100
     idle-timeout = 30 s
-    # Based on throughput of 50 req/sec * 10 seconds for recovery
-    queue-length = 500
+    # Based on throughput of 50 req/sec * 5 seconds for recovery
+    queue-length = 250
   }
 
   log4j {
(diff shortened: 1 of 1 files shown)


soda-fountain-lib/src/main/resources/reference.conf
```diff
@@ -92,8 +92,8 @@ com.socrata.soda-fountain = {
     min-threads = 10
     max-threads = 100
     idle-timeout = 30 s
-    # Based on throughput of 50 req/sec * 60 seconds for recovery
-    queue-length = 3000
+    # Based on throughput of 50 req/sec * 10 seconds for recovery
```
kaidamasaki: Where does 50 req/sec come from?

soda-fountain-lib/src/main/resources/reference.conf
```diff
@@ -92,8 +92,8 @@ com.socrata.soda-fountain = {
...
+    queue-length = 500
```
znep: I might set it even lower... it seems this resembles http://en.wikipedia.org/wiki/Bufferbloat in a lot of ways at a higher protcol level.  We don't want requests backing up at every layer of the stack, it is better to push queuing as far upstream as possible to avoid unnecessary extra concurrency all the way down the stack.

soda-fountain-lib/src/main/resources/reference.conf
velvia: Profiling of sustainable load in SEA1.

soda-fountain-lib/src/main/resources/reference.conf
velvia: That's fine.  Where is the queue upstream though?  Core?    Do we need to do something about a queue in Core?

velvia:
@znep halved it.

PR #94, opened 2015-04-02
add suggest sample endpoint

sokrahta:
+ plumbing to route suggest sample requests from core to spandex

[Review on Reviewable](https://reviewable.io/reviews/socrata/soda-fountain/94)

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Suggest.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Suggest.scala
@@ -6,6 +6,7 @@ import javax.servlet.http.HttpServletResponse
 import com.rojoma.json.v3.ast.JNull
 import com.socrata.http.client._
 import com.socrata.http.client.exceptions.ContentTypeException
+import com.socrata.http.server._
 import com.socrata.http.server.implicits._
 import com.socrata.http.server.responses._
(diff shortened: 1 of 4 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Suggest.scala
```diff
@@ -44,41 +45,52 @@ case class Suggest(datasetDao: DatasetDAO, columnDao: ColumnDAO,
...
-          val body = try {
-            spandexResponse.jValue()
-          } catch {
-            case e: ContentTypeException => log.warn(s"Non JSON response: $e")
-              JNull
-          }
+  case class sampleService(resourceName: ResourceName, columnName: ColumnName) extends SodaResource {
+    override def get = { req => resp =>
+      log.info(s"GET /suggest $resourceName :: $columnName [sample]")
+      go(resourceName, columnName, (dataset, copynum, column, _) =>
```
velvia: I don't understand .... so the get method returns a function  HttpResource => URI ?

sokrahta: the get method returns `HttpService` which is alias to `HttpRequest` => `HttpServletResponse`

velvia:
Ok, lgtm.

PR #102, opened 2015-04-11
Random cleanup

velvia:
- Delete the soda-fountain-war project
- Make 'assembly' much faster by only building assembly for the jetty project (and root project- not sure of a way to disable it - but the root project is empty)
- turn two very verbose logging messages down to debug
- add metrics for suggest and sample routes
- more debugging for SoQLPackDumper


velvia:
@chitang I don't see "earldouglas" or anything like that in project/plugins.sbt....

PR #105, opened 2015-04-27
Improve /healthz endpoint to provide more useful info when it fails

urmilan:
CORE-5227

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/clients/geospace/GeospaceClient.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/clients/geospace/GeospaceClient.scala
@@ -1,26 +1,54 @@
 package com.socrata.soda.clients.geospace
 
+import com.socrata.soda.clients.geospace.GeospaceClient._
+import com.socrata.soda.server.config.GeospaceClientConfig
 import com.socrata.thirdparty.curator.CuratorServiceBase
 import org.apache.curator.x.discovery.ServiceDiscovery
-import scala.util.Try
-import scalaj.http.Http
+import scalaj.http.{HttpOptions, Http}
+import com.rojoma.json.v3.util.AutomaticJsonCodecBuilder
 
-trait GeospaceClient {
-  def versionCheck: Boolean
+object GeospaceClient {
+  val PartialResponseLength = 500
+
+  sealed trait VersionCheckResult
+  case object Success extends VersionCheckResult
+  case class Failure(url: String, status: Int, partialResponse: String) extends VersionCheckResult
+  object Failure {
+    implicit val jCodec = AutomaticJsonCodecBuilder[Failure]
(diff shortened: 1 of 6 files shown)


urmilan:
@znep Let me know if there is anything else useful to include.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/HealthZ.scala
```diff
@@ -4,17 +4,23 @@ import com.rojoma.json.v3.ast.{JBoolean, JObject}
...
 
   object service extends SodaResource {
     override def get = { req => response =>
-      val geospaceOk = geospace.versionCheck
-      val status = if (geospaceOk) OK else InternalServerError
+      val (code, fields) = geospace.versionCheck match {
+        case Success =>
+          (OK, Map("geospace_ok" -> JBoolean.canonicalTrue))
+        case f: Failure =>
+          val details = JsonEncode.toJValue(f)
```
znep: Can we also log the failure reason?  Unless we are doing that somewhere I missed.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/geospace/GeospaceClient.scala
```diff
@@ -1,26 +1,54 @@
...
+import com.socrata.soda.server.config.GeospaceClientConfig
 import com.socrata.thirdparty.curator.CuratorServiceBase
 import org.apache.curator.x.discovery.ServiceDiscovery
-import scala.util.Try
-import scalaj.http.Http
+import scalaj.http.{HttpOptions, Http}
+import com.rojoma.json.v3.util.AutomaticJsonCodecBuilder
+
+object GeospaceClient {
+  val PartialResponseLength = 500
```
znep: This is relatively obvious, but maybe just a short comment about what this value is especially since the number 500 may make people think "internal server error?  huh"?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/HealthZ.scala
urmilan: Addressed in efdd045

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/geospace/GeospaceClient.scala
urmilan: Addressed in efdd045

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/HealthZ.scala
urmilan: Actually in 53d9ad7

PR #108, opened 2015-05-06
Speed up GeoJSON export

velvia:
- using the Grisu Double toString serializer
- using optimizations in third-party-utils GeoJSON encoder

GeoJSON export should be roughly an order of magnitude faster.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -17,13 +17,14 @@ object Dependencies {
     val postgresql        = "9.1-901-1.jdbc4"
     val quasiQuotes       = "2.0.0"
     val rojomaJson        = "3.2.2"
+    val rojomaJsonGrisu   = "1.0.0"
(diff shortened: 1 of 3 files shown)


urmilan:
Can you investigate the [test failure](http://jenkins.aws-us-west-2-infrastructure.socrata.net/job/sbt-build/530/console)? Apart from that lgtm.

project/Dependencies.scala
```diff
@@ -17,13 +17,14 @@ object Dependencies {
     val postgresql        = "9.1-901-1.jdbc4"
     val quasiQuotes       = "2.0.0"
     val rojomaJson        = "3.2.2"
```
urmilan: Do you still need this?

velvia: Not really but might be good to keep around to easily switch back.

sokrahta:
lgtm :shipit:

PR #109, opened 2015-05-15
Parse WKT values on ingest (enables CSV geom ingest)

velvia:
Also enables us to ingest what is exported from NBE via CSV.   :)


velvia:
@chitang No, soda-fountain really only ingests JSON.  Core right now translates CSV to JSON for soda fountain.

We should think about adding real CSV ingest to soda-fountain.

PR #110, opened 2015-05-15
Refs CORE-5412

rozap:
* added multipoint polygon and line types
* added soql tests
* TODO: update the dependencies file and remove the snapshot reference

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/GeoJsonExporter.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/GeoJsonExporter.scala
@@ -84,11 +84,11 @@ class GeoJsonProcessor(writer: BufferedWriter, schema: ExportDAO.CSchema, single
 
     soqlGeom match {
       case SoQLPoint(p)         => geoCodec.encode(p)
-      case SoQLMultiLine(ml)    => geoCodec.encode(ml)
-      case SoQLMultiPolygon(mp) => geoCodec.encode(mp)
-      case SoQLLine(l)          => geoCodec.encode(l)
       case SoQLMultiPoint(mp)   => geoCodec.encode(mp)
+      case SoQLLine(l)          => geoCodec.encode(l)
+      case SoQLMultiLine(ml)    => geoCodec.encode(ml)
       case SoQLPolygon(p)       => geoCodec.encode(p)
+      case SoQLMultiPolygon(mp) => geoCodec.encode(mp)
       case _                    => JNull
     }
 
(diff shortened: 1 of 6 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/CsvColumnRep.scala
```diff
@@ -2,7 +2,7 @@ package com.socrata.soda.server.wiremodels
 
 import com.rojoma.json.v3.io.CompactJsonWriter
 import com.socrata.soql.types._
-import com.vividsolutions.jts.geom.{Geometry, MultiLineString, MultiPolygon, Point}
+import com.vividsolutions.jts.geom.{Geometry, MultiLineString, MultiPolygon, Point, LineString, Polygon, MultiPoint}
```
urmilan: I wonder if it's just worth `com.vividsolutions.jts.geom._` at this point

soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/GeoJsonExporter.scala
```diff
@@ -81,10 +81,14 @@ class GeoJsonProcessor(writer: BufferedWriter, schema: ExportDAO.CSchema, single
...
 
   private def getGeometryJson(soqlGeom: SoQLValue): JValue =
+
     soqlGeom match {
       case SoQLPoint(p)         => geoCodec.encode(p)
       case SoQLMultiLine(ml)    => geoCodec.encode(ml)
       case SoQLMultiPolygon(mp) => geoCodec.encode(mp)
+      case SoQLLine(l)          => geoCodec.encode(l)
+      case SoQLMultiPoint(mp)   => geoCodec.encode(mp)
+      case SoQLPolygon(p)       => geoCodec.encode(p)
```
urmilan: Let's keep the existing convention and list the geometries in order of complexity: Point, Line, Polygon, Multipoint, Multiline, Multipolygon. Here and elsewhere

urmilan:
LGTM. If we can think of a way to make the GeometryLikeRep stuff less boilerplate, that would be great. I think I tussled with that a bit when I initially added the other geo types and didn't come up with a good solution. Robert or Millstone might have ideas.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/GeoJsonExporter.scala
urmilan: Or Point, MultiPoint, Line, MultiLine, Polygon, Multipolygon, which I see is what you ahve done in data-coordinator. Just some kind of logical order. :)

rozap:
Yea, I discussed the `GeometryLikeRep` ugliness with @kaidamasaki  and @rjmac and sort of agreed that it was unpleasant but the alternatives where also ugly. :(

urmilan:
:shipit: Feel free to merge after replacing the snapshot dependencies.

PR #111, opened 2015-05-19
Use region-coder service instead of Geospace for region coding

urmilan:


sokrahta:
looks fine to me; 
there aren't any dangling geospace references in soda-fountain
tests are passing
:shipit:

PR #113, opened 2015-05-21
Add JMX port to Docker config

velvia:


rozap:
not familiar with JMX but this looks reasonable

velvia:
@rozap  You might want to read up on it....  the standard network interface for remote and local debugging of the JVM, live control of apps, profiling, etc.

rozap:
doing so atm :+1:

PR #115, opened 2015-06-11
Krah/spandex error message

sokrahta:
I thought maybe we should give a better error message than "null" when soda-fountain couldn't get a spandex response for some reasons.


urmilan:
We should be consistent with the rest of soda-fountain and use `SodaUtils.errorResponse` (should be plenty of examples elsewhere in the project).

sokrahta:
no need for a new error case class, I'll just leverage InternalException

sokrahta:
actually InternalException isn't cooperating with the socrata-http exceptions, so I'm creating a new case class

urmilan:
:shipit:

PR #116, opened 2015-06-12
Run metasoda migrations when the container starts.

urmilan:
Most of our migrations are ideally run _before_ a new version of a service is running. In some cases, not running the migrations beforehand can lead to outages. 

It's not uncommon that we forget to run migrations at the appropriate time to each environment; this change makes sure that never happens. 

The upshot of this is that if we have a migration that needs to be run *after* changes to the service, we'd need to deploy the service code change and the migration in 2 separate steps.

### DIFF
--- a/docker/ship.d/run
+++ b/docker/ship.d/run
@@ -3,7 +3,7 @@ set -ev
 
 /bin/env_parse ${SERVER_CONFIG}.j2
 
-exec su socrata -c '/usr/bin/java \
+su socrata -c '/usr/bin/java \
     -Dconfig.file=${SERVER_CONFIG} \
     -cp $SERVER_ARTIFACT \
     com.socrata.soda.server.MigrateSchema migrate'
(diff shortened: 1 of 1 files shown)


docker/ship.d/run
```diff
@@ -4,6 +4,11 @@ set -ev
 /bin/env_parse ${SERVER_CONFIG}.j2
 
 exec su socrata -c '/usr/bin/java \
```
znep: Does this actually still run soda fountain?  It seems like the exec here would result in not running the actual service...

velvia:
What about just adding an option to have the main soda fountain runtime run migrations when it starts?  Then it works in any environment, including SEA1...

velvia:
and is more convenient arguably for local testing too.

docker/ship.d/run
sokrahta: yes it runs migration process first, then continues on to the main process

sokrahta:
SEA1 will be gone soon enough to ignore, maybe just an update to the bin/start_soda_fountain.sh script

sokrahta:
SEA1 will be gone soon enough.
There is `bin/run_migrations.sh` maybe just consolidate that with `bin/start_soda_fountain/sh`

PR #117, opened 2015-06-22
Disallow source columns from being deleted without removing the dependency first

urmilan:
The lack of validation for this scenario has resulted in some hairy situations in production - let's just prevent it from happening in the first place.


urmilan:
Thank @chitang, please see the latest changes.

urmilan:
Fixed in 61ac39e

PR #122, opened 2015-07-30
Don't generate multiple request IDs in a single request.

urmilan:
Basically we were trying to pull out the same header from an HttpRequest twice. Changed the code to pull it out once and subsequently pass it between methods.

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
@@ -97,13 +97,13 @@ case class Resource(rowDAO: RowDAO,
 
   def upsertishFlow(req: HttpServletRequest,
                     response: HttpServletResponse,
+                    requestId: RequestId.RequestId,
                     resourceName: ResourceName,
                     rows: Iterator[JValue],
                     f: rowDaoFunc) = {
     datasetDAO.getDataset(resourceName, None) match {
       case DatasetDAO.Found(datasetRecord) =>
-        val transformer = new RowDataTranslator(
-          RequestId.getFromRequest(req), datasetRecord, false)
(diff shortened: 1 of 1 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
```diff
@@ -97,13 +97,13 @@ case class Resource(rowDAO: RowDAO,
...
   def upsertishFlow(req: HttpServletRequest,
                     response: HttpServletResponse,
+                    requestId: RequestId.RequestId,
                     resourceName: ResourceName,
                     rows: Iterator[JValue],
                     f: rowDaoFunc) = {
     datasetDAO.getDataset(resourceName, None) match {
       case DatasetDAO.Found(datasetRecord) =>
-        val transformer = new RowDataTranslator(
-          RequestId.getFromRequest(req), datasetRecord, false)
```
kaidamasaki: Was this called multiple times for a single request?  

This seems like something we should potentially fix upstream, though I'm not sure if it could be done easily.

urmilan: Yes, see lines 222 and 227.

urmilan: And agreed that it should go upstream somewhere, but I want to make sure that this is the crux of the upsert request ID logging issues first.

kaidamasaki: Cool, glad you tracked down the root.  

I'm worried that we have this issue in other places, potentially, is why I think potentially we would want to have something in the library that keeps track of if a request id has been generated for a request, it might be overly complicated to keep a thread local somewhere per request, though.

PR #124, opened 2015-08-20
passing dc errors

chitang:
error handling is complicated and do not work well.  This is not a complete solution to the problem.  I hope this is an improvement and should have better coverage.


urmilan:
:shipit:

PR #125, opened 2015-08-26
Update region coding computation handlers to use region-coder v2 endpoints

urmilan:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/computation/GeoregionMatchOnPointHandler.scala
```diff
@@ -37,7 +37,9 @@ class GeoregionMatchOnPointHandler[T](config: Config, discovery: ServiceDiscover
       case Some(ComputationStrategyRecord(_, _, _, Some(JObject(map)))) =>
```
rozap: not related to this changeset, but what is map? could it be named something more descriptive than "map"? because we know it's a map.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/computation/GeoregionMatchOnStringHandler.scala
```diff
@@ -38,7 +38,9 @@ class GeoregionMatchOnStringHandler[T](config: Config, discovery: ServiceDiscove
         require(map.contains("column"), "parameters does not contain 'column'")
         val JString(region) = map("region")
         val JString(column) = map("column")
```
rozap: this too :confused:

rozap:
lgtm :shipit:

PR #126, opened 2015-08-28
Refs CORE-5975

rozap:
Add a GET request query parameter to the /export endpoint
which will explicitly specify the CRS of the export.
Not going to name names, but this is needed
for certain libraries that crash when you parse geoJSON
which follows the spec correctly and includes the CRS.
Currently only crs84 is supported, and that is all that
will be supported because anything else will break the
geoJSON spec.


rozap:
nevermind, this hack won't be necessary because i'm planning to abandon geotools

PR #129, opened 2015-09-21
Make computed columns writable for upcoming async changes

urmilan:


rozap:
lgtm, i like deletions. but i don't really understand how the changes accomplish the goal here, i short description might enlighten me. either way, :shipit:

urmilan:
Right now, Soda Fountain calculates computed column values and bars any external caller from writing those values. This just removes that safeguard so that the new async region jobs can take over calculating computed column values.

PR #131, opened 2015-09-28
Add SoQLBlob support

rjmac:
It's pretty trivial since there's nothing soda-fountain can do with these except pass them through.  They aren't even eligible to be primary keys (which is why it's commented out in `StringColumnRep`).


meganschneider:
:shipit:

PR #134, opened 2015-10-14
Add option include_system_columns=[true|false] in export.

chitang:


urmilan:
To avoid confusing ourselves later, can we deprecate and remove the existing systems column list at lines 101-111 of `ColumnSpecUtils.scala` as part of this commit?

chitang:
Lines 101-111, the explicit list of system columns is still used.  And the details it needs is more than a simple system column test (prefix ":").  I think that at least belong to a separate refactor.

chitang:
Hi Urmila, after you are done.  Please don't merge because I want to squash these.

PR #135, opened 2015-10-14
Remove unnecessarily noisy table dropper log messages

urmilan:
This log message currently spams the soda fountain log every 30 seconds - removing it.


rozap:
huzzah! :ship:

PR #136, opened 2015-10-14
Fix dataset lookup bug

urmilan:
Subtle bug in the dataset lookup SQL query that prevents the most recent dataset copies from being returned if there have been any discarded working copy on a dataset. This prevents any subsequent working copies from being created.


znep:
:ship: 

The only comment I'd make is it may be worth renaming `dcDeletedFilter` to something ugly but more verbose like `latestCopyDeletedFilter` to avoid the confusion of `dc`.  Going even further would alias `dataset_copies` in the subselect to something like `latest_dc` then you could qualify those bits with that.

May not be worth actually making the change at this point, but that is what I'd do when writing the query to reduce the chances of seeing the bug in the first place... or that is what I tell myself in hindsight.

urmilan:
See latest commit - I tried to make the SQL query a little clearer.

znep:
:+1:

PR #137, opened 2015-10-15
Add option exclude_system_fields=[true|false] in export.

chitang:
Copy SoQLPackWriter from SoQLReference since this is the only place using it.


chitang:
Change my mind.  I am going to leave SoQLPackWriter in SoQL reference.

PR #138, opened 2015-10-15
Add option exclude_system_fields=[true|false] in export.

chitang:


chitang:
Wait...structural type reflective performance is actually worse than creating a new array.  Not going to use it.

PR #144, opened 2015-10-30
Expose unsupported column type creation errors. 

AEscobarCruz:
added response parameter to withColumnSpec in order to respond appropriately with error responses instead of an empty 200.


meganschneider:
:shipit:

PR #146, opened 2015-11-26
Help mimic soda1 by making row id obfuscation optional.

chitang:


chitang:
I'll change the query parameter name from id_appearance to obfuscate_id next week to be consistent.

aerust:
Looks good!

PR #147, opened 2015-12-16
Cleanup some internal error responses

aerust:


saraid:
i think this does not look terrible and also the demo worked so :shipit:

PR #148, opened 2015-12-18
EN-158 - adds geocoding computation handler to Soda-fountain

aerust:
Cassandra configuration and mapquest credentials will need to be added to soda-fountain config.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -25,7 +25,7 @@ object Dependencies {
     val socrataGeocoders  = "1.0.6"
     val socrataHttp       = "3.3.3"
     val soqlBrita         = "1.3.0"
-    val soqlReference     = "0.8.9"
+    val soqlReference     = "0.8.10"
     val thirdPartyUtils   = "4.0.10"
     val curatorUtils      = "1.0.3"
     val typesafeConfig    = "1.0.2"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -24,7 +24,7 @@ object Dependencies {
     val socrataHttp       = "3.3.3"
     val soqlBrita         = "1.3.0"
     val soqlReference     = "0.8.9"
-    val thirdPartyUtils   = "3.1.2"
+    val thirdPartyUtils   = "4.0.5"
     val curatorUtils      = "1.0.3"
     val typesafeConfig    = "1.0.2"
 
@@ -78,7 +78,7 @@ object Dependencies {
   val socrataHttpJetty         = "com.socrata"     %% "socrata-http-jetty"          % versions.socrataHttp
   val socrataHttpServer        = "com.socrata"     %% "socrata-http-server"         % versions.socrataHttp
 
-  val socrataThirdPartyUtils   = "com.socrata"     %% "socrata-thirdparty-utils"    % "4.0.5"
+  val socrataThirdPartyUtils   = "com.socrata"     %% "socrata-thirdparty-utils"    % versions.thirdPartyUtils
   val socrataCuratorUtils      = "com.socrata"     %% "socrata-curator-utils"       % versions.curatorUtils
 
   val soqlBrita                = "com.socrata"     %% "soql-brita"                  % versions.soqlBrita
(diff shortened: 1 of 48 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/SodaFountain.scala
```diff
@@ -141,33 +139,100 @@ class SodaFountain(config: SodaFountainConfig) extends Closeable {
...
+      val cache = config.cache match {
+      case Some(cacheConfig) =>
+        astyanax match {
+          case Some(ast) =>
+            new CassandraCacheClient (ast.getClient, cacheConfig.columnFamily, cacheConfig.ttl)
+          case None =>
+                log.info("No Cassandra configuration provided; using NoopCacheClient for geocoding.")
+                NoopCacheClient
+            }
+          case None =>
```
rjmac: The indentation here's a little weird.  If you're not attuned to the braces it'll look like there are two "case None"s for the inner match.

project/Dependencies.scala
```diff
@@ -37,68 +39,74 @@ object Dependencies {
     val wiremock          = "1.46"
   }
 
-  val apacheCuratorDiscovery   = "org.apache.curator" % "curator-x-discovery"   % versions.apacheCurator
+  def astyanaxExcludes(x: ModuleID) = x exclude ("commons-logging", "commons-logging") exclude ("org.mortbay.jetty", "servlet-api") exclude ("javax.servlet", "servlet-api")
+  val astyanaxCassandra = astyanaxExcludes("com.netflix.astyanax" % "astyanax-cassandra" % versions.astyanaxVersion)
+  val astyanaxThrift    = astyanaxExcludes("com.netflix.astyanax" % "astyanax-thrift"    % versions.astyanaxVersion)
```
rjmac: Poor diff algorithm gets so confused from the whitespace changes....

soda-fountain-lib/src/main/scala/com/socrata/soda/server/SodaFountain.scala
```diff
@@ -77,19 +82,12 @@ class SodaFountain(config: SodaFountainConfig) extends Closeable {
   val rng = new scala.util.Random(new SecureRandom())
   val columnSpecUtils = new ColumnSpecUtils(rng)
 
-  private val cleanup = new mutable.Stack[Closeable]
+  private val resourceScope = parentResourceScope.open(new ResourceScope("soda-fountain"))
```
rjmac: Might want to add a note about what we discussed re: this being only an approximation to prompt cleanup.  Or not, as also discussed it'll be prompt in practice.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/computation/GeocodingHandler.scala
```diff
@@ -0,0 +1,226 @@
...
+   * @param computedColumn Computed column definition
+   * @return               Map from address component to some column id if it is defined
+   */
+  private def getColumnIds(computedColumn: ColumnRecordLike): Map[String, Option[ColumnId]] = {
+    val fieldNames = getColumnNames(computedColumn)
+    computedColumn.computationStrategy match {
+      case Some(ComputationStrategyRecord(_, _, Some(sourceColumns), _)) =>
+        fieldNames.mapValues {
+          case Some(name) =>
+            sourceColumns.find(column => column.fieldName.equals(name)) match {
```
rjmac: Prefer `==` to `equals` in scala code, unless you're doing something very unusual.  It's annoying, going back and forth with Java, but `==` reads better.  (In fact, I'd probably write that as `_.fieldName == name` instead of `column => column.fieldName == name`)

soda-fountain-lib/src/main/scala/com/socrata/soda/server/computation/GeocodingHandler.scala
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Geocodes a batch of addresses to points
+   * @param addresses Sequence of addresses to geocode
+   * @return          Sequence of results; a result will be Some(Point) if the address
+   *                  can be meaningfully geocoded, else the result will be None
+   */
+  private def geocode(requestId: RequestId, addresses: Seq[Address]): Seq[Option[Point]] = {
+    try {
+      print(addresses)
```
rjmac: stray `print`s

PR #153, opened 2016-01-16
Refs EN-1737

rozap:
* bump logging up


vilterp:
:shipit: as long as this won't flood our logs!

PR #156, opened 2016-01-19
Assign explicit port (6012) for the liveness check

rozap:
* Also update misc things to work with bumped
  version of socrata-http library

Refs EN-1737


bspeakmon:
LGTM.

PR #157, opened 2016-01-20
BUG: EN-1783 - rollup soql should not be chained.

chitang:
The final rollup soql saved in pg should not be wrapped by LIST(soql).


rjmac:
Will this allow a user to cause a 500 error?

chitang:
Let's see how it can be turned into an user error.

PR #158, opened 2016-01-23
Fix test break due to socrata-http version bump.

chitang:
Mock http requests need to expect get request id call.


NathalieKaligirwa:
lgtm ship it

PR #159, opened 2016-02-10
additional logic added to export only certain columns

AEscobarCruz:
Should help in performing region coding much more efficiently and with less memory constraints than loading full rows to memory.

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Export.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Export.scala
@@ -9,14 +9,14 @@ import com.socrata.http.server.implicits._
 import com.socrata.http.server.responses._
 import com.socrata.http.server.routing.OptionallyTypedPathComponent
 import com.socrata.http.server.util.{Precondition, EntityTag, RequestId}
-import com.socrata.soda.server.SodaUtils
+import com.socrata.soda.server.copy.Stage
 import com.socrata.soda.server.errors._
 import com.socrata.soda.server.export.Exporter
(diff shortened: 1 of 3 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Export.scala
```diff
@@ -82,6 +82,34 @@ case class Export(exportDAO: ExportDAO, etagObfuscator: ETagObfuscator) {
...
+    // get only these columns, we expect these to be fieldnames comma separated items.
+    val columnsOnly = Option(req.getParameter("columns")).map {
+      paramStr =>
+        try {
+          val columnFields = paramStr.toLowerCase.split(",").toSeq.map(x => x.trim)
+          val columns = exportDAO.lookupColumns(resourceName, Stage(copy)).getOrElse(Seq.empty)
+          // need to have the row-identifier included otherwise dc will fail
+          // this may seem long winded but helps avoid double counting in case one of requested
+          // columns is the row-identifier
+          val rowIdColumn = columns.filter{c => c.typ == SoQLID}.map{c => c.fieldName.name}
```
NathalieKaligirwa: Is SoQLID, system column `:id`?

AEscobarCruz: can be, i believe and any columned labeled as a row-identifier.

AEscobarCruz:
The store object is not available at the export level only in the doa level should it even be exposed, no?

chitang:
That's right.  :shipit:

PR #160, opened 2016-02-14
Chi/en 101/export by rowid

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorErrors.scala
```diff
@@ -46,6 +46,9 @@ case class ContentTypeUnknownCharset(`content-type`: String) extends DCRequestEr
 @Tag("req.export.mismatched-schema")
 case class SchemaMismatchForExport(dataset: DatasetId, schema: SchemaSpec) extends DCRequestError
 
+@Tag("req.export.invalid-row-id")
+case class InvalidRowId() extends DataCoordinatorError
```
AEscobarCruz: I think this is a DCRequestError, if not please move this to the Single error case section.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Export.scala
```diff
@@ -134,6 +134,8 @@ case class Export(exportDAO: ExportDAO, etagObfuscator: ETagObfuscator) {
       case other => return SodaUtils.errorResponse(req, BadParameter("sorted", other))(resp)
     }.getOrElse(true)
 
+    val rowId = Option(req.getParameter("row_id"))
```
AEscobarCruz: anything we can catch right away? like empty request parameter sent?

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorErrors.scala
chitang: Yes, moved to DCRequestError.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Export.scala
chitang: empty row_id is OK since this is optional.  Deeper check requires more work that I rather not repeat in DC like figuring out copy, custom row id and column rep.

AEscobarCruz:
:shipit:

PR #163, opened 2016-02-19
Return whole row if insert payload is single row { ... }

chitang:
curl -v -X POST --header "Content-Type: application/json" "http://localhost:6010/resource/_8dsk-iq93" --data '{"code":"aa","speed": 1}'

{":created_at":"2016-02-18T17:16:32.624Z",":id":"row-gm3b~q4ku-jhpf",":updated_at":"2016-02-19T17:06:13.913Z",":version":"rv-su34-c89g-49vi","speed":"1","code":"aa"}

curl -v -X POST --header "Content-Type: application/json" "http://localhost:6010/resource/_8dsk-iq93" --data '[{"code":"aa","speed": 1}]'

[{"typ":"update","id":"aa","ver":"rv-57q5~2bc3.jbxp"}]


soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/InputUtils.scala
```diff
@@ -56,9 +56,11 @@ object InputUtils {
    * into an iterator with a single jvalue.  This allows processing of multiple rows and single row.
    * @note The iterator can throw a `TooMuchDataWithoutAcknowledgement` exception if the user
    *       sends an element which cannot be read within `approximateMaxDatumBound` bytes.
+   * @return Two items, jvalue iterator and a boolean indicating whether the input is
+   *         in brackets (true) or in braces (false)
    */
   def jsonArrayValuesStream(req: HttpRequest, approximateMaxDatumBound: Long, allowSingleItem: Boolean)
-    : Either[SodaError, Iterator[JValue]] = {
+    : Either[SodaError, Tuple2[Iterator[JValue], Boolean]] = {
```
AEscobarCruz: This change did not have other dep. that needed changing?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/UpsertUtils.scala
```diff
@@ -95,4 +99,52 @@ object UpsertUtils {
...
     }
   }
+
+  def writeSingleRowUpsertResponse(resourceName: ResourceName, export: Export, req: HttpRequest)
+                                  (response: HttpServletResponse, report: Iterator[ReportItem]): Unit = {
+    var wroteOne = false
+    while(report.hasNext) {
+      report.next() match {
+        case UpsertReportItem(items) =>
+          while(items.hasNext) {
```
AEscobarCruz: how about instead of looping you call the first item, check if there is another still and from there log one error with a list of remaining items if you must? Similar situation to the above with report.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Export.scala
```diff
@@ -136,6 +139,37 @@ case class Export(exportDAO: ExportDAO, etagObfuscator: ETagObfuscator) {
...
+               limit,
+               offset,
+               columnsOnly,
+               excludeSystemFields,
+               ifModifiedSince,
+               sorted,
+               rowId: Option[String], false)(req)(resp)
+  }
+
+  def exportCopy(resourceName: ResourceName, copy: String, ext: Option[String],
```
AEscobarCruz: perhaps you can get away without overloading by having a default singleRow = false value.

AEscobarCruz:
for the most part lgtm, just a lot of layers as always in soda.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/InputUtils.scala
chitang: No, only called from one place.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Export.scala
chitang: One form takes many export arguments from the request while the other form has arguments not from the request.

PR #164, opened 2016-02-23
Wasn't passing excludeSystemFields downstream.

chitang:


meganschneider:
:shipit: confirmed fixed locally too

PR #165, opened 2016-03-01
Exporting with computed column fails when exclude_system_fields = true

chitang:
We assumed that all system columns clumps together which does not hold for computed columns.  Now that DC export does support specifying columns, SF will use that to exclude system columns.


AEscobarCruz:
just awesome! :shipit:

PR #169, opened 2016-03-05
EN-3523: When adding a computed column, clean up the orphaned column …

urmilan:
…if computation fails


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/DatasetColumn.scala
```diff
@@ -103,16 +103,23 @@ case class DatasetColumn(columnDAO: ColumnDAO, exportDAO: ExportDAO, rowDAO: Row
     }
 
     override def put = { req => resp =>
+      val userFromReq = user(req)
+      val requestId = RequestId.getFromRequest(req)
```
urmilan: I remember running into an issue a while ago with these methods not necessarily being idempotent, so that's why I'm calling them just once here.

bjornbytes:
:shipit:

PR #170, opened 2016-03-08
Use latest soql-reference for type check simplification.

chitang:
There isn't an immediate need.  Just to keep project up-to-date.


rjmac:
lgtm

PR #172, opened 2016-03-08
aerust/EN-3506/part-2

aerust:
Drop recompute column only after all instances of
soda-fountain no longer reference the column.


JoeNunnelley:
@aerust no changes?

aerust:
@JoeNunnelley @chitang , yeah essentially no changes, and to be in a release following part one. This is to not cause down time for eu-west: https://socrata.atlassian.net/browse/EN-3506

aerust:
This commit should not be merged until commit c1553f284b9fa4e6a6ee3b20a79b04722617c539 is released

PR #174, opened 2016-03-09
EN-3506: Drop recompute column safely (part 1.25)

aerust:
Add default value for recompute column, so references
to it can be removed in the future.


aerust:
Okay put it directly into the xml file, I'm going to force push because it's already such a small commit.

PR #175, opened 2016-03-09
Simplest possible geojson support for dataset with multiple geo columns

urmilan:
When originally implementing geojson, we completely locked down geojson export on datasets with multiple geo columns, however we've had a decent number of requests to support this scenario and it seems to make sense to loosen these restrictions somewhat.

As an initial step, the exporter picks the first geo column it happens to encounter as the primary feature geometry. Other geometry column values will be relegated to the attributes section of the feature. 

Ideally we'd allow the API caller to pass a parameter indicating which column should be the primary geometry, but that's a bigger refactor that we haven't prioritized yet.

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/GeoJsonExporter.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/GeoJsonExporter.scala
@@ -24,15 +24,10 @@ object GeoJsonExporter extends Exporter {
   val mimeType = new MimeType(mimeTypeBase)
   val extension = Some("geojson")
 
-  // For now, GeoJSON only works if you have exactly ONE geo column in the dataset.
+  // For now, GeoJSON only works if you have at least ONE geo column in the dataset.
   // Attempting to export a dataset with zero or more than one geo columns will return HTTP 406.
-  // For non-geo datasets the reason for this behavior is obvious.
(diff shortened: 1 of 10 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/GeoJsonExporter.scala
```diff
@@ -69,16 +58,22 @@ class GeoJsonProcessor(writer: BufferedWriter, schema: ExportDAO.CSchema, single
 
   val jsonWriter = new CompactJsonWriter(writer)
 
-  private def getGeoColumnIndex(columns: Seq[ColumnInfo]): Int = {
+  private def getGeoColumnIndex(columns: Seq[ColumnInfo]): Option[Int] = {
     val geoColumnIndices = columns.zipWithIndex.collect {
       case (columnInfo, index) if columnInfo.typ.isInstanceOf[SoQLGeometryLike[_]] => index
     }
 
-    // We validate the dataset schema upfront and throw a 406 on datasets that don't
```
urmilan: I wrote the original geojson export so this is me.

AEscobarCruz:
+1 Again for fixing things.
LGTM and thank you for writing notes on areas about what the decision was on for this change. Still let's talk to @rozap about this change before shipping it.

PR #177, opened 2016-03-11
Use the right resource scope when querying

rjmac:
So responses don't get cut off in the middle.


cyranix:
Is there any test we can add to guard against regression?

AEscobarCruz:
@cyranix pretty sure there is a cheetah test that could be written out.

rjmac:
Cheetah is the right place for that, yeah.

AEscobarCruz:
probably should assign that as a ticket for your group. @chitang, thoughts?

AEscobarCruz:
:shipit: tested it locally, as soon as pending tests are done, think we can roll this out.

chitang:
Yes, nbego team should create a test to cover that.  Joe has been working with larger data.  He may already have something close to ready.

PR #178, opened 2016-03-15
En 172

rjmac:
Set up the new NBE internal shapshot APIs for finding, listing, exporting, and deleting snapshots from truth.


meganschneider:
:shipit:

PR #179, opened 2016-03-16
EN-3886: Secondary versions of dataset

aerust:
Add /dataset-version[resourceName] endpoint
which returns
 - the version of the dataset in truth
 - the versions that secondaries have of dataset
 - set of feedback secondaries

-- Requires commit in data-coordinator:
2c58d04ad479bd2ce9c8227c0596f0d4fe9d9cb9
--


aerust:
Pending checks:
- Needs testing
- data-coordinator pr

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/HttpDataCoordinatorClient.scala
```diff
@@ -383,6 +384,28 @@ abstract class HttpDataCoordinatorClient(httpClient: HttpClient) extends DataCoo
...
+        .get
+      httpClient.execute(request).run { response =>
+        response.resultCode match {
+          case HttpServletResponse.SC_OK =>
+            val oVers = response.value[SecondaryVersionsReport]()
+            oVers match {
+              case Right(vers) => vers
+              case Left(_) => throw new Exception("secondary versions not found")
+            }
+          case HttpServletResponse.SC_NOT_FOUND => throw new Exception("dataset not found")
```
rjmac: Not a fan of these exceptions; these feel like they should be "normal" cases for this method to return.

aerust: Okay, I was kind of thinking that myself... will go make those changes

PR #181, opened 2016-03-22
Allow this location form:

chitang:
{ "human_address": {"city": "seattle" }, "latitude": 5.5, "longitude": 6.6 } 

in addition to:

{ "human_address": "{\"city\": \"seattle\" }", "latitude": 5.5, "longitude": 6.6 } 

to match obe capability.


rjmac:
Ok.  I'm actually surprised it accepts the latter and not the former now...

chitang:
I thought the later is more "official".  It is the default output format (soda2 v2 may have changed that) as well as how it is physically stored.

PR #182, opened 2016-03-24
En 172

rjmac:
More exporting of snapshots


meganschneider:
:shipit:  still lgtm ;)

PR #184, opened 2016-04-06
EN-2679: Do not send soda-fountain metrics to AMQ

urmilan:
The main urgency around this ticket is to reduce load on ActiveMQ. Given there was some ambiguity around whether we want to rip out metrics completely or send them to Graphite, this commit simply redirects all soda-fountain metrics to the `NoopMetricProvider` for now. Once Slemko is back from vacation we can discuss if we want to redirect those metrics to Graphite.


AEscobarCruz:
lgtm, could you put a TODO there that we need to revisit?

urmilan:
Will do

AEscobarCruz:
:shipit: when TODO is written. Thanks!

PR #185, opened 2016-04-11
Pass request header X-Socrata-Fuse-Columns to query-coordinator.

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/RowDAOImpl.scala
```diff
@@ -116,11 +118,13 @@ class RowDAOImpl(store: NameAndSchemaStore, dc: DataCoordinatorClient, qc: Query
...
                       obfuscateId: Boolean,
                       requestId: RequestId,
+                      fuseColumns: Option[String],
                       resourceScope: ResourceScope): Result = {
     val extraHeaders = Map(ReqIdHeader              -> requestId,
                            SodaUtils.ResourceHeader -> ds.resourceName.name,
                            "X-SODA2-DataVersion"    -> ds.truthVersion.toString,
-                           "X-SODA2-LastModified"   -> ds.lastModified.toHttpDate)
+                           "X-SODA2-LastModified"   -> ds.lastModified.toHttpDate) ++
+      fuseColumns.map(c => Map("X-Socrata-Fuse-Columns" -> c)).getOrElse(Map.empty)
```
AEscobarCruz: So you map each column to a map object that has a key that maps to that column...? Just want to make sure this is correct?

chitang: Not exactly what it does here.  It is just changing an optional string to a request header.  The string is in this form -> field1,fuse_rule1;field2,fuse_rule2;...  Eventually, this string will turn into a map.  But that doesn't happen until it gets to Query Coordinator where the actual work is performed.

AEscobarCruz:
LGTM, :shipit:

PR #187, opened 2016-04-18
Add Url datatype.

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/JsonColumnRep.scala
```diff
@@ -98,6 +98,25 @@ object JsonColumnRep {
     }
   }
 
+  object UrlRep extends JsonColumnRep {
+    val representedType = SoQLUrl
+
+    def fromJValue(input: JValue): Option[SoQLValue] = {
+      JsonDecode[SoQLUrl].decode(input) match {
+        case Right(x: SoQLUrl) => Some(x)
```
rjmac: Shouldn't need the type annotation here...

rjmac:
lgtm

PR #188, opened 2016-04-21
EN 4781: Undelete API

TranquilMarmot:


TranquilMarmot:
This goes with https://github.com/socrata/core/pull/280

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Dataset.scala
```diff
@@ -139,6 +141,12 @@ case class Dataset(datasetDAO: DatasetDAO, maxDatumSize: Int) {
     }
   }
 
+  case class undeleteService(resourceName: ResourceName) extends SodaResource {
+    override def post = { req =>
```
TranquilMarmot: I wanted to make this a `patch` but it complained about not having any data when called 😢

soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/PostgresStoreImpl.scala
```diff
@@ -595,6 +595,31 @@ class PostgresStoreImpl(dataSource: DataSource) extends NameAndSchemaStore {
...
+      val datasetIdOpt = using(connection.prepareStatement("select dataset_system_id from datasets where resource_name_casefolded = ? and deleted_at is not null for update")) { idFetcher =>
+        idFetcher.setString(1, resourceName.caseFolded)
+        using(idFetcher.executeQuery()) { rs =>
+          if (rs.next()) Some(DatasetId(rs.getString(1)))
+          else None
+        }
+      }
+      for (datasetId <- datasetIdOpt) {
+        using(connection.prepareStatement(
+          """ update dataset_copies dc set  deleted_at = null where dc.dataset_system_id = ?;
```
chitang: Can we just undelete all copies?  May be we need to look up from truth to see what copies are available and only undelete those.

I don't know how/when columns info are deleted in SF.  Is there anything need to be done?

TranquilMarmot: Truth == user db, right? How would we go about looking them up there? There definitely needs to be some sort of mechanism to prevent undeleting a dataset that has been 'purged' from the system.

As for columns being deleted; this undelete operation is supposed to be something that only admins can do in a worst-case kind of scenario. I think there are some expected side effects of undeleting a dataset, like deleted columns coming back.

meganschneider: No. Truth is truth (NBE data only, original). Userdbs are OBE-only.

meganschneider: Also yeah, I suspect that all of the columns entries for the datasets get hard-deleted when the dataset/copy is initially soft-deleted. Also computation_strategies.

chitang: After looking at what delete dataset does, I am not concerned about checking what copies are available in truth any more.  It is a bonus if we have a general dataset integrity check that can be called any time including after undelete.  I would add the condition "deleted_at is not null" to the update sqls.  It doesn't do anything normally.

chitang: It would be nice to keep the line under 120 chars and format the SQLs nicely.

chitang:
lgtm.  just want better formatting so that the line is within 120 chars.  Also need some unit tests.

PR #189, opened 2016-04-22
EN-3506: Drop recompute column safely (part 2)

aerust:
Drop recompute column only after all instances of
soda-fountain no longer reference the column.

-- Must be in a seperate release from commit
c1553f284b9fa4e6a6ee3b20a79b04722617c539
--


rjmac:
ok

aerust:
Robert says "👍 "

PR #190, opened 2016-04-29
EN-3709: allow single-row upsert with custom row ids which are integers

meganschneider:
Previously this failed with upsert-error-malformed-report-item-id
because these end up as JNumbers, not JStrings. Dates appear to be taken
care of as strings already.

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/UpsertUtils.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/UpsertUtils.scala
@@ -103,6 +103,16 @@ object UpsertUtils {
   def writeSingleRowUpsertResponse(resourceName: ResourceName, export: Export, req: HttpRequest)
                                   (response: HttpServletResponse, report: Iterator[ReportItem]): Unit = {
     var wroteOne = false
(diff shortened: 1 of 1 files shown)


aerust:
lgtm

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/UpsertUtils.scala
```diff
@@ -103,6 +103,16 @@ object UpsertUtils {
   def writeSingleRowUpsertResponse(resourceName: ResourceName, export: Export, req: HttpRequest)
                                   (response: HttpServletResponse, report: Iterator[ReportItem]): Unit = {
     var wroteOne = false
+    def exportSingleRowUpsertResponse(rowId: String): Unit = {
+      val param = ExportParam(None, None, Seq.empty[ColumnRecordLike], None,
```
JoeNunnelley: can you align the parameter list with the first parameter when they span multiple lines. it helps readability. thanks

meganschneider: Fixed!

JoeNunnelley:
:shipit:

PR #192, opened 2016-05-17
EN-5016: make soda-fountain unaware of column descriptions/human names

meganschneider:
this is the soda-fountain side of a core fix; these are not
interdependent and either one will fix most of the cases in the bug
(although both are required for the case of updating both fieldname and
description)


rjmac:
:+1:

PR #193, opened 2016-05-19
EN-4817: Validate geocoding computation strategies

aerust:
Also transform `computationStrategy.parameters` to
have column ids instead of field names

Refs EN-3522


aerust:
I want to add some tests around what I have added.

aerust:
Force pushed after rebasing onto master

rjmac:
:+1:

PR #195, opened 2016-06-02
Acquiring locks timeouts in 30s for migration.

chitang:


rjmac:
Will this roll back _all_ migrations that would be run, or only the changeset that failed to acquire the lock?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/NonCommittingJdbcConnection.scala
```diff
@@ -0,0 +1,41 @@
...
+    ct += 1
+    for (stmt <- managed(conn.createStatement())) {
+      stmt.execute(s"SAVEPOINT x$ct")
+    }
+  }
+
+  override def rollback(): Unit = {
+    if (ct > 0) {
+      for (stmt <- managed(conn.createStatement())) {
+        stmt.execute(s"SAVEPOINT x$ct")
```
rjmac: ROLLBACK TO SAVEPOINT, yes?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/NonCommittingJdbcConnection.scala
```diff
@@ -0,0 +1,41 @@
...
+  * When this connection is used for liquibase migration, changeSet.runInTransaction must be true(which is the default).
+  */
+class NonCommmittingJdbcConnenction(conn: Connection) extends JdbcConnection(conn) {
+
+  private var ct = 0
+
+  override def commit(): Unit = {
+    ct += 1
+    for (stmt <- managed(conn.createStatement())) {
+      stmt.execute(s"SAVEPOINT x$ct")
```
rjmac: I feel like it should also establish a savepoint at initial construction, so LB can do "sqlstuffsqlstuff; rollback" correctly.  As it is until at least one commit has happened there'll be no savepoint to roll back to.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/NonCommittingJdbcConnection.scala
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+
+  override def rollback(): Unit = {
+    if (ct > 0) {
+      for (stmt <- managed(conn.createStatement())) {
+        stmt.execute(s"ROLLBACK TO SAVEPOINT x$ct")
+      }
+      ct -= 1
```
chitang: It really isn't meant to replace normal commit/rollback logic.  It is just to get around the problem where LB always rollbacks semantic which isn't a problem to them because they already commit at the end of a changeSet.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/NonCommittingJdbcConnection.scala
chitang: This probably explains why it "works" even when "rollback to" was missing.  It can really be anything valid thing except rollback.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/NonCommittingJdbcConnection.scala
rjmac: It should be possible to do committish semantics though.  Something like (pseuocodish):

```scala
  var savepoint = conn.setSavepoint()
  override def commit(): Unit = {
    conn.releaseSavepoint(savepoint)
    savepoint = conn.setSavepoint()
  }
  override def rollback(): Unit = {
    conn.rollback(savepoint)
    savepoint = conn.setSavepoint()
  }
  def realCommit(): Unit = {
    conn.releaseSavepoint(savepoint)
    conn.commit()
  }
```

rjmac:
heh, ok, that's even simpler

PR #198, opened 2016-06-10
EN-6972: Not all non-200 responses are errors!

znep:
We keep logging things like "Error  responding with error not-modified:
Resource was not modified".  This offense my delicate sensibilities
as this isn't an error and we shouldn't say it is.  So rename the use of
"error" to be "response".

When reviewing, ensure nothing is actually changing in the HTTP
response.  All changes should be internal and only expressed in how
things are logged to avoid compatibility issues.


saraid:
Main concern is that "success" is also a type of response. But seems fine to me. :shipit:

PR #199, opened 2016-06-16
EN-5555: divide soda exception classes into internal errors and invalid

meganschneider:
requests

Then at the top level, handle these appropriately with 4xx and 5xx
responses and logging.
Also fix one spot where we were erroneously throwing
CannotDeletePrimaryKey instead of DeleteWithoutPrimaryKey.


aerust:
lgtm

aerust:
What testing do we want to do around this?

meganschneider:
Some of this will get asserted by cheetah tests (nbe failures); trying to sort out the rest. Fewer cheetah tests are failing against my local than I'd expect right now.

aerust:
Sounds reasonable enough.

PR #201, opened 2016-06-20
EN-5555: handle integer custom row ids when given wrong ids

meganschneider:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/id/AbstractId.scala
```diff
@@ -49,6 +49,7 @@ object RowSpecifier {
     def encode(x: RowSpecifier): JValue = JString(x.underlying)
     def decode(x: JValue): DecodeResult[RowSpecifier] = x match {
       case JString(s) => Right(RowSpecifier(s))
+      case n: JNumber => Right(RowSpecifier(n.toString()))
```
chitang: no need for ()

meganschneider: intellij was giving me a warning when I remove them, not sure why.

chitang:
lgtm

PR #202, opened 2016-06-25
Relay invalid life cycle stage error.

chitang:
This error happens when you ask for creating working copy which already exists.

The return code is 409.  There was also typo/inconsistent use of stage and state which caused decoding problem.

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorErrors.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorErrors.scala
@@ -130,8 +130,8 @@ case class CannotAcquireDatasetWriteLock(dataset: DatasetId) extends DCDatasetUp
 
 @Tag("update.dataset.invalid-state")
 case class IncorrectLifecycleStage(dataset: DatasetId,
-                                   `actual-stage`: String,
-                                   `expected-stage`: Set[String]) extends DCDatasetUpdateError
+                                   `actual-state`: String,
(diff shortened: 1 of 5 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorErrors.scala
```diff
@@ -130,8 +130,8 @@ case class CannotAcquireDatasetWriteLock(dataset: DatasetId) extends DCDatasetUp
 
 @Tag("update.dataset.invalid-state")
 case class IncorrectLifecycleStage(dataset: DatasetId,
-                                   `actual-stage`: String,
-                                   `expected-stage`: Set[String]) extends DCDatasetUpdateError
+                                   `actual-state`: String,
```
meganschneider: shouldn't this also be stage, given that the thing that is incorrect is a stage? And the tag as well? Guessing that side is in data-coordinator though..

chitang: Yes/No.  The is the last piece I changed in this PR after adding all the missing pieces and couldn't get this to work e2e.  This problem lies in DC - DatasetUpdateError.INVALID_STATE = "update.dataset.invalid-state"  Keeping this stage requires another set of changes in DC which I prefer not to touch.  I prefer to confine changes to SF alone.

meganschneider: Fair enough. :shipit:

PR #203, opened 2016-06-28
Switch to turn off region-coding in soda-fountain

rjmac:
Controlled via zookeeper; defaults to "region-coding is enabled".


aerust:
Looks good to me :boat:

PR #205, opened 2016-07-25
EN-6194 support single row in PUT resource/4x4

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
```diff
@@ -266,7 +266,7 @@ case class Resource(rowDAO: RowDAO,
 
     override def put = { req => response =>
       val requestId = RequestId.getFromRequest(req)
-      upsertMany(req, response, requestId, rowDAO.replace(user(req), _, _, requestId), allowSingleItem = false)
+      upsertMany(req, response, requestId, rowDAO.replace(user(req), _, _, requestId), allowSingleItem = true)
```
meganschneider: I have no idea what else this may affect.

chitang: I think this only affects putting to /resource/4x4.  What I don't recall is if there was a reason for allowSingleItem = false.  I don't think there was.  I was probably just focusing on post when I added single item and just default everything else to the original value.

rjmac: Does it have the right semantics?  In particular: does it NOT delete the entire dataset?

IIRC, single-row PUT on update is supposed to replace the row completely (including replacing unsupplied values with NULL) and single-row POST is supposed to replace those columns which are explicitly provided only.

PR #214, opened 2016-10-05
EN-5578:  Expand `/dataset-version/_4x4` endpoint

aerust:
Changes to `/dataset-version/_4x4` endpoint:
 * `truth_version` is now optional
 * adds `latest` data-version
 * adds `published` data-version
 * adds `unpublished` data-version
 * adds `groups` - a map from `group_name` to the set of
   `store_id`s (only the `store_id`s that the
    dataset is replicated to)

All new fields are optional.


chitang:
:ship:

chitang:
lgtm

aerust:
Cloudbees failure looks like a silly jenkins error... not going to worry about it.

PR #217, opened 2016-10-30
remove cloudbees references

JoeNunnelley:
@urmilan please review and build. FYI: It appears that building has errors currently (before my change).


JoeNunnelley:
we've decided to abandon this work for now.

PR #219, opened 2016-12-06
EN-12392 missing charset in content-type header

chitang:


aerust:
Looks like this breaks the unit tests on cloudbees? Once they pass looks good to me

aerust:
lgtm :shell:

PR #220, opened 2017-01-10
EN-12821: Add temporary extra logging to dataset-version endpoint to …

urmilan:
…investigate customer issues


aerust:
lgtm :sheep:

PR #221, opened 2017-01-17
EN-12140 Also write X-SODA2-Fields in csv export.

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/Exporter.scala
```diff
@@ -20,7 +23,33 @@ trait Exporter {
   protected def exporterHeaders(schema: ExportDAO.CSchema): HttpResponse =
     schema.lastModified.fold(NoOp) { lm =>
       Header("Last-Modified", HttpUtils.HttpDateFormat.print(lm))
+      maybeSoda2FieldsHeader(schema)
     }
+
+  /**
+    * Override in specific exporters if we want X-SODA2-Fields
+    */
```
aerust: I think I'm just missing it here, but what does the caller have to do to get the X-SODA2-Fields?

aerust:
I think added a bit of an explanation to the commit message of _why_ you are making this change would be good.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/Exporter.scala
chitang: Override mayBeSoda2FieldsHeader to call writeSoda2FieldsHeader.  That is done in JsonExporter and TableExporter.

aerust: But will it be always that they are added?

chitang: There are other unique exporters (cjson, soqlpack...) in nbe that I still don't want to give in.

aerust: Oh, okay. :+1:

aerust:
Commit message is nice :) looks good to me :sheep:

PR #222, opened 2017-02-02
Don't throw away last-modified header from truth

rjmac:


chitang:
:shipit: thank you for fixing this

PR #223, opened 2017-02-03
Support multiple data coordinators

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/SodaFountain.scala
```diff
@@ -175,7 +177,9 @@ class SodaFountain(config: SodaFountainConfig) extends Closeable {
     new FeedbackSecondaryManifestClient(dc, feedbackSecondaryIdMap)
   }
 
-  val datasetDAO = i(new DatasetDAOImpl(dc, fbm, store, columnSpecUtils, () => config.dataCoordinatorClient.instance))
+  val datasetDAO = i(new DatasetDAOImpl(
+    dc, fbm, store, columnSpecUtils,
+    () => config.dataCoordinatorClient.instances(rng.nextInt(config.dataCoordinatorClient.instances.size))))
   val columnDAO = i(new ColumnDAOImpl(dc, fbm, store, columnSpecUtils))
```
aerust: I guess I'm slightly confused on how this is suppose to work: if you have a pre-existent dataset you want to pick the instance that the dataset is on, if you are creating a dataset you want to pick some (random) instance. Isn't this always returning a random instance?

Other thought: random seems like a fine start for now for how to pick an instance. Do we have any idea how or when we might do something more sophisticated?

znep:
As far as picking an instance for new datasets, the thing it seems we would
want to start based on where we are now is always going to a specific one
(ie. the new one).

chitang:
"instances" means instances for creation only.  pre-existing datasets request should automatically go to the appropriate dc instance because of curated http client.  If we only wants new datasets to go to one new instance only, instances is set to [ "bravo" ].  If we have two brand new truths, we may want to set it to [ "bravo", "charlie" ].  Since "alpha" is near full capacity, we don't include it here.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/config/SodaFountainConfig.scala
```diff
@@ -30,8 +31,15 @@ class SodaFountainConfig(config: Config) extends ConfigClass(WithDefaultAddress(
 class DataCoordinatorClientConfig(config: Config, root: String) extends ConfigClass(config, root) {
   val serviceName = getString("service-name")
   val instance = getString("instance")
```
znep: Can we make this private or otherwise not expose it as part of the DataCoordinatorClientConfig, to ensure nothing can use it by accident?

Hrm, looking further we do use it at https://github.com/socrata-platform/soda-fountain/blob/fad3e2ebff3915a4e1079ef4098192de0e815a61/soda-fountain-lib/src/main/scala/com/socrata/soda/server/SodaFountain.scala#L145-L145 ... how does that work?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/config/SodaFountainConfig.scala
```diff
@@ -30,8 +31,15 @@ class SodaFountainConfig(config: Config) extends ConfigClass(WithDefaultAddress(
...
-  val connectTimeout = getDuration("connect-timeout")
-  val receiveTimeout = getDuration("receive-timeout")
+  val instances =
+    try {
+      getStringList("instances").toVector
```
znep: Can we name this something more obvious than instances?  eg. `instances-for-new-datasets` or something similar.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/config/SodaFountainConfig.scala
chitang: I think it is a property that doesn't get used.

PR #226, opened 2017-03-09
EN-123901: Don't deadlock when region coding fails

aerust:
When computing a newly created computed column, if
computation fails, then close the export from
data-coordinator before trying to drop the column
from truth.


aerust:
I don't have the unit tests passing yet... need to figure out how to fix them.

PR #229, opened 2017-03-31
EN-15144: Pass resource_name to data-coordinator

aerust:
Pass on dataset `resource_name` in mutation script
to data-coordinator upon dataset creation. This is
so secondary-watcher can send replication complete
events with dataset external ids for waiting for
computing column statistics.


chitang:
lgtm

PR #100, opened 2015-04-08
Check the result set in fetchMinimalColumn before trying to read things out of it

urmilan:
If a source column in computation_strategy becomes invalid for whatever reason, Soda Fountain currently blows up trying to read its (non-existent) column definition (see stack trace below). Fixing that.

```
2015-04-07 16:28:45,998 ERROR [6078 / DELETE /dataset/_wfqv-uxqr/%3A%40computed_region_mv22_r2ri] (2a282bfbe72b3be687a292596efcedec) [] Error Internal exception: 0810c56d-cf9d-45c3-8a72-321f67acbe4d
org.postgresql.util.PSQLException: ResultSet not positioned properly, perhaps you need to call next.
	at org.postgresql.jdbc2.AbstractJdbc2ResultSet.checkResultSet(AbstractJdbc2ResultSet.java:2675)
	at org.postgresql.jdbc2.AbstractJdbc2ResultSet.getString(AbstractJdbc2ResultSet.java:1852)
	at org.postgresql.jdbc2.AbstractJdbc2ResultSet.getString(AbstractJdbc2ResultSet.java:2385)
	at com.mchange.v2.c3p0.impl.NewProxyResultSet.getString(NewProxyResultSet.java:3326)
	at com.socrata.soda.server.persistence.pg.PostgresStoreImpl.parseMinimalColumn(PostgresStoreImpl.scala:344)
	at com.socrata.soda.server.persistence.pg.PostgresStoreImpl$$anonfun$fetchMinimalColumn$2$$anonfun$apply$48.apply(PostgresStoreImpl.scala:337)
...
```

PR #230, opened 2017-04-07
EN-14870 Add date_extract_xxx(floating_timestamp) functions. (#100)

chitang:
These functions provide support for the apps team.

All is needed here is bump soql-lib.  This only matters for rollups.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -24,7 +24,7 @@ object Dependencies {
     val scalaj            = "0.3.15"
     val socrataHttp       = "3.10.1"
     val soqlBrita         = "1.3.0"
-    val soqlReference     = "2.0.11"
+    val soqlReference     = "2.1.1"
     val thirdPartyUtils   = "4.0.5"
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -24,7 +24,7 @@ object Dependencies {
     val scalaj            = "0.3.15"
     val socrataHttp       = "3.10.1"
     val soqlBrita         = "1.3.0"
-    val soqlReference     = "2.0.11"
+    val soqlReference     = "2.1.1"
```
aerust: Why was soda-fountain behind on soql-reference compared to core?

aerust:
lgtm

project/Dependencies.scala
chitang: Normally, SoQLs pass through in string form w/o being analyzed except when we create rollups.  Sometimes I don't bump version in SF if they aren't expected to be used in rollups.
That middle digit jump was for distinct.

PR #232, opened 2017-05-09
EN-16161 - Add Zookeeper Whitelist for Secondary Geo- and Region-Coding

cmlaidlaw:
- Refactor ComputingGate to accept a resource name and then check if it exists in a whitelist (stored in Zookeeper) before deciding whether or not to send the geo- or region-coding job to the secondary as opposed to doing it in Soda Fountain itself.
- We are doing this so that we can selectively test the correctness and stability of the new geo- and region-coding secondary on a dataset-by-dataset basis before turning it on globally per environment.
- We expect the entire ComputingGate mechanism to be removed at the same time that we disable Soda Fountain's geo- and region-coding functionality.

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/computation/ComputingGate.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/computation/ComputingGate.scala
@@ -1,26 +1,72 @@
 package com.socrata.soda.server.computation
 
+import com.socrata.soda.server.id.ResourceName
+
 import java.io.Closeable
 import java.nio.charset.StandardCharsets
 
 import org.apache.curator.framework.CuratorFramework
 import org.apache.curator.framework.recipes.cache.NodeCache
 
-class ComputingGate(curator: CuratorFramework, path: String) extends (() => Boolean) with Closeable {
-  val nc = new NodeCache(curator, path + "/computation-enabled")
+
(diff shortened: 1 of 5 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/computation/ComputingGate.scala
```diff
@@ -1,26 +1,45 @@
...
+    val computationWhitelistData = computationWhitelistNodeCache.getCurrentData
+
+    val computationEnabledInEnvironment =
+      computationEnabledInEnvironmentData != null &&
+        !java.util.Arrays.equals(computationEnabledInEnvironmentData.getData, falseAsBytes)
+    val resourceNameWhitelist = Option(computationWhitelistData.getData) match {
+      case Some(resourceWhitelistData) => new String(resourceWhitelistData, StandardCharsets.UTF_8).
+        split(',').
+        toSeq.
+        map(new ResourceName(_))
```
aerust: I think I might just also call the .contains here and not return Nil, but return false. But that might be just because I've never used Nil before.

aerust:
So, once this done, I think you should merge this to master and we'll test it quickly in staging. Then we'll checkout a branch off of the last release version, cherry pick this, tag a release on that branch (that should build our docker image), then merge that branch so Joe can tag a release on soda-fountain without conflict tomorrow.

aerust:
Looks great :goat: :boat: 🎏

PR #237, opened 2017-07-12
EN-17534: Don't do work when computation is off

aerust:
Don't do extra work when computation has been
disabled for a given dataset. The uneeded upserting
of the source column done by soda-fountain is
causing geocoding secondary to get 409s while
it is trying to region code large datasets.


aerust:
So the function ComputeUtils.compute is called by both paths you can compute via soda-fountain.

chitang:
:shipit:

PR #238, opened 2017-07-12
EN-17537: Don't check for deleted tables to drop insanely frequently.

znep:
And add index on datasets(deleted_at).  Looking up datasets to delete was the 
second most expensive query hitting soda-fountain by CPU.


znep:
No, I've already pre-run it in fedramp and sf doesn't have long running
transactions normally.

aerust:
Awesome :+1:

PR #239, opened 2017-07-13
EN-17534: Fix bug in 73dfa0c6ecdfd5e1930b369e5a9dcdb9c04322a8

aerust:
Call the success handler when skipping computation when computation
is not enabled for a dataset. So soda-fountain can return the correct
response when a computed column is created. Also note this means
when the /compute endpoint is used and the computing gate is off
it will say it computed 0 rows (which is true).


chitang:
:shipit:

PR #244, opened 2017-08-18
A very basic backup plan to collocate dataset on soda2 query.

chitang:
This is at least useful for development

curl -H ‘X-Socrata-Collocate: true’ ‘http://sf:6010/resource/4x4?$query=select * join @x on @x.c=c’

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
@@ -7,8 +7,7 @@ import com.socrata.soda.clients.datacoordinator.DataCoordinatorClient.ReportItem
 
 import scala.collection.JavaConverters._
 import scala.language.existentials
-
-import com.rojoma.json.v3.ast.{JNumber, JString, JValue}
+import com.rojoma.json.v3.ast.{JArray, JNumber, JString, JValue}
 import com.rojoma.simplearm.util._
 import com.rojoma.simplearm.v2.ResourceScope
 import com.socrata.http.common.util.ContentNegotiation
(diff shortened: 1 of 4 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
```diff
@@ -233,6 +233,9 @@ case class Resource(rowDAO: RowDAO,
                     SodaUtils.response(req, SodaErrors.RequestTimedOut(timeout))
                   case RowDAO.QCError(status, qcErr) =>
                     metricByStatus(status)
+                    if (Option(req.getHeader("X-Socrata-Collocate")) == Some("true")) {
+                      collocateDataset(qcErr)
```
aerust: So, I'm kind of scared on only relying on this... since we have no way of preventing this from happening?

chitang: We have not exposed this in core-server.  It should be inaccessible.

aerust: Oh, okay, I guess, how is it not exposed in core-server?

chitang:
core only pass through specific headers.

aerust:
Okay, I guess I'm kind of worried about it making a mess... maybe we can add a zookeeper flag so we can turn it on/off?

aerust:
Though, I suppose it will probably be only you using it /shrug

chitang:
It will be only me using it until we decide that we are not investing to implement the full fledged version in a timely manner :).

aerust:
Okay, I'll let you merge it, but only to make you happy :) :P

chitang:
no need for zk flag unless we expose it in core.

PR #246, opened 2017-09-21
I didn’t get the secondary-id right.

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
```diff
@@ -303,7 +303,7 @@ case class Resource(rowDAO: RowDAO,
               datasetDAO.getDataset(resourceName, None) match {
                 case DatasetDAO.Found(datasetRecord) =>
                   val dsId = datasetRecord.systemId
-                  val secId = SecondaryId(sec.split("\\.").head)
+                  val secId = SecondaryId(sec.split("\\.")(1))
                   log.info("collocate {} dataset {} in {}", resourceName, dsId.underlying, secId)
```
aerust: okay

PR #252, opened 2017-10-20
EN-19514: Remove region coding computation

aerust:
Remove region coding computation paths from Soda-fountain.

Leave the /compute endpoint and add some logging which I think is
unused until we are confident that it is safe to get rid of.


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Dataset.scala
```diff
@@ -8,7 +8,7 @@ import com.socrata.http.server.util.RequestId
 import com.socrata.soda.server._
 import com.socrata.soda.server.copy.Stage
 import com.socrata.soda.server.responses._
-import com.socrata.soda.server.highlevel.DatasetDAO.{NonExistentColumn, DatasetAlreadyExists}
+//import com.socrata.soda.server.highlevel.DatasetDAO.{NonExistentColumn, DatasetAlreadyExists}
```
chitang: You don't mean to check this in?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Dataset.scala
```diff
@@ -8,7 +8,7 @@ import com.socrata.http.server.util.RequestId
...
 import com.socrata.soda.server.highlevel._
```
aerust: oops I'll delete this

chitang:
Compute isn't really doing anything even though you leave the endpoint there?

chitang:
🚢 red is good

PR #254, opened 2017-10-31
EN-19703 Add SoQLDocument type

chitang:
Add SoQL Photo type.


soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/CsvColumnRep.scala
```diff
@@ -173,6 +175,41 @@ object CsvColumnRep {
     }
   }
```
aerust: Can we make a point to delete the CSV code path out of soda-fountain, since we don't want to support it for the document type?

PR #255, opened 2017-11-14
EN-19155: bump soqlReference version to 2.9.1 to support CollectionExtract fns

bahrieinn:
To support changes here: https://github.com/socrata-platform/soql-reference/pull/92


chitang:
:ship:

PR #256, opened 2017-11-20
EN-19576 Increase soda-fountain jetty max request size for datasets w…

chitang:
…ith lots of columns


znep:
Did you validate that this is sufficient and we don't need to tweak any other params too?  IIRC seeing comments about there also being another limit on request size or some such you start running into...

If this is all that is required, 🚢

chitang:
No, I did not validate.  Last time we had to update socrata-http to allow configuration.  Since it isn't urgent, I plan to verify this fix when it is deployed.

PR #257, opened 2017-12-28
EN-20581: setup soda-fountain for local container dev

michaelb990:


.solo/soda-fountain.json
```diff
@@ -0,0 +1,6 @@
+{
+    "aliases": [ "sf" ],
+    "start_process": "bin/start_soda_fountain.sh",
+    "start_container": "bin/start_soda_fountain_docker.sh",
+    "ports": [ 6010 ]
```
tkbrigham: nit: can we get `ports` below `aliases` for alphabetic sorting?

soda-fountain-lib/src/main/resources/reference.conf
```diff
@@ -9,10 +9,16 @@ com.socrata.soda-fountain = {
...
     name = "soda-fountain"
+    address = "local.dev.socrata.net"
   }
 
   database {
     app-name = "soda-fountain"
+    host = "local.dev.socrata.net"
+    port = 5432
+    username = "blist"
+    password = "blist"
```
tkbrigham: assuming this is ok since it's just for dev stuff?

bin/start_soda_fountain_docker.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+image_tag="$1"
+if [ "$image_tag" == "local" ]; then
```
tkbrigham: Do we want any sort of sane default that grabs the most recent if no `$1` is present?

.solo/soda-fountain.json
```diff
@@ -0,0 +1,6 @@
+{
+  "aliases": [ "sf" ],
+  "ports": [ 6010 ],
```
tkbrigham: 👍

bin/start_soda_fountain_docker.sh
michaelb990: not sure what that would be. the `solo` script has the logic to find the latest tag on the registry. i don't really want to duplicate that in these too. what would you want to default it to?

soda-fountain-lib/src/main/resources/reference.conf
michaelb990: yes

bin/start_soda_fountain_docker.sh
tkbrigham: That's a fair point. I wasn't sure how easy it would be to identify the latest as a fallback, like if there were a single call to just grab the latest image. But if it involves hitting some endpoints and adding logic, I agree that it's not appropriate.

michaelb990: I'll add some error handling though.

soda-fountain-lib/src/main/resources/reference.conf
znep: Just seeing this now but ... ideally we have tried to keep localhost and internal socrata specific dev config out of the reference.conf for public NBE projects, although people have slipped a few in.

PR #261, opened 2018-01-10
EN-16556: Collocation Routing

PiSoup:
EN-16556: Implemented route through soda-fountain for collocation. This only handles transform of resources->ids, data-coordinator is responsible for business logic around collocation.


PiSoup:
Error handling in this is mostly just a stub. It will need to be properly fleshed out for https://socrata.atlassian.net/browse/EN-21382

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorClient.scala
```diff
@@ -45,6 +46,7 @@ object DataCoordinatorClient {
   // SUCCESS CASES
   case class NonCreateScriptResult(report: Iterator[ReportItem], etag: Option[EntityTag], copyNumber: Long, newVersion: Long, lastModified: DateTime) extends SuccessResult
   case class ExportResult(json: Iterator[JValue], lastModified: Option[DateTime], etag: Option[EntityTag]) extends SuccessResult
+  case class ColocateResult(status: String, message: String) extends SuccessResult
```
aerust: for consistency `CollocateResult` should have two `l`'s.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/HttpDataCoordinatorClient.scala
```diff
@@ -605,6 +616,45 @@ abstract class HttpDataCoordinatorClient(httpClient: HttpClient) extends DataCoo
...
       }
     }
+
+  override def collocate(secondaryId: SecondaryId, operation: DCCollocateOperation): Result = {
+    implicit val encoder = AutomaticJsonEncodeBuilder[DCCollocateOperation]
+
+    // Use any of the dcs mentioned in the operation as a host
+    withHost(operation.collocations.head.head) { host =>
+      val request = collocateUrl(host)
+        .addParameter("explain" -> "true")
```
aerust: Do you want explain hard coded to true for now? (fine if we do)

aerust: But maybe it is worth parsing that parameter in soda-fountain and passing it on?

PiSoup: We're going to need to wire it up through core as well. I'll at least handle the parse in soda-fountain for the moment though and default it to explain=true.

PR #264, opened 2018-01-31
EN-21909: Passing JobID for Collocation

PiSoup:
Added jobID parameter to collocate endpoint for use by data-coordinator. Soda-fountain now returns full status received from data-coordinator. This makes the soda layer behave a bit more transparently for collocations.


PiSoup:
This strikes me as a ridiculous amount of code to write to create a transparent layer. Anyone have a suggestion for making this a bit more organized?

PR #266, opened 2018-02-22
EN-17026: Pass data-coordinator collocation limits

aerust:
Pass on cost limits to data-coordinator so data-coordinator
may reject collocation if the cost exceeds the passed limits.

Warning: this is a breaking change since the new fields in the
collocation request are required. Soda-fountain should be deployed
prior to data-coordinator to not cause errors.


soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorClient.scala
```diff
@@ -7,7 +7,7 @@ import com.socrata.soda.server.id._
 import com.socrata.soda.server.persistence.ColumnRecord
 import com.socrata.soda.server.util.CopySpecifier
 import com.socrata.soda.server.util.schema.SchemaSpec
-import com.socrata.http.server.util.{Precondition, EntityTag}
```
PiSoup: Intellij! :shakefist:

PR #267, opened 2018-02-23
EN-22593: Route collocation status endpoint

aerust:
Route collocation status endpoint through Soda-fountain.

This adds the new route:
`GET /dataset-copy/{resourceName}/collocate/{secondaryId}/job/{jobId}`

Also, translate dataset system ids coming from Data-coordinator in the
response to resource names and add some new fields.


soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/PostgresStoreImpl.scala
```diff
@@ -126,6 +126,30 @@ class PostgresStoreImpl(dataSource: DataSource) extends NameAndSchemaStore {
     }
   }
 
+  def translateDatasetId(datasetId: DatasetId, isDeleted: Boolean = false): Option[ResourceName] = {
```
PiSoup: Could you just use `bulkDatasetLookup` instead? It does a `Set[DatasetId] => Set[ResourceName]`

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/HttpDataCoordinatorClient.scala
```diff
@@ -618,39 +619,40 @@ abstract class HttpDataCoordinatorClient(httpClient: HttpClient) extends DataCoo
...
+    case class CollocateResponse(id: Option[String],
+                                 status: String,
+                                 message: String,
+                                 cost: Cost,
+                                 moves: Seq[Move])
+
+    implicit val cCodec = AutomaticJsonCodecBuilder[CollocateResponse]
+
+    response.value[CollocateResponse]() match {
+      // NOTE: data-coordinator response has the field "id" where core expects "jobId"
```
PiSoup: Think it's worth getting every service using the same name or no? I'm on the fence about it but if we're going to, now is the time.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAO.scala
```diff
@@ -81,6 +95,17 @@ object DatasetDAO {
...
   case object RollupDropped extends SuccessResult
   case class CollocateDone(jobId : Option[String], status: String, message: String, cost: Cost, moves: Seq[Move]) extends SuccessResult
+  object CollocateDone {
+    def apply(r: DataCoordinatorClient.CollocateResult, translator: DatasetId => Option[ResourceName]): CollocateDone = {
+      CollocateDone(
+        r.jobId,
+        r.status,
+        r.message,
+        Cost(r.cost),
+        r.moves.flatMap(Move(_, translator)) // TODO: this will filter out datasets that have been deleted (not yet seen by truth), not sure what the right thing to do is here?
```
PiSoup: This probably makes the most sense. If a dataset doesn't exist here then we don't really want this service doing anything to investigate. If we want to act on that information the frontend will have enough to recognize the discrepancy and alert the user at the job status page.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/PostgresStoreImpl.scala
aerust: Oh cool, I didn't notice that function.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/HttpDataCoordinatorClient.scala
aerust: I think I would rather leave the "id" field in data-coordinator how it is to be consistent with naming conventions there.

PiSoup:
:shipit:

PR #277, opened 2018-10-06
EN-27677: Pass through 429/503 status code.

znep:
This is to allow for the load shedding per dataset concurrency limit
changes in https://github.com/socrata-platform/soql-postgres-adapter/pull/284 to
pass a reasonable HTTP status code back.

This is mostly sidestepping the whole complicated and unfinished
soda fountain error code infrastructure in favor of standardized
HTTP status codes.

We only need 429 for the load shedding work, but I implemented 503
first before I changed my mind and went with 429, and think 503 is
valuable to pass back in general instead of the previous behavior
of turning the 503 into a generic 500 if it didn't have a json
response body.


chitang:
:shipit: so much boilerplate

PR #278, opened 2018-10-10
EN-27755 Support add/drop computation strategy on existing columns

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/ColumnDAOImpl.scala
```diff
@@ -226,22 +226,49 @@ class ColumnDAOImpl(dc: DataCoordinatorClient,
...
+              store.addComputationStrategy(datasetRecord.systemId, copyNumber, spec.copy(id = columnRecord.id))
+              instructionsBuilder += AddComputationStrategyInstruction(columnRecord.id, cs)
+            }
+          }
+          if (spec.computationStrategy.isEmpty && columnRecord.computationStrategy.nonEmpty) {
+            store.dropComputationStrategy(datasetRecord.systemId, copyNumber, spec.copy(id = columnRecord.id))
+            instructionsBuilder += DropComputationStrategyInstruction(columnRecord.id)
+          }
+        case x =>
+          log.warn("columnSpecUtils.freezeForCreate(update) failed and ignored", x)
```
PiSoup: Why is this only a warn?

chitang: This runs for all column updates.  Historically, we have been ignoring computation strategy stuff.  I don't want this to suddenly throw error on requests that may not care about computation strategy.

PR #283, opened 2018-12-05
EN-28370 Able to create RowUpdateOptionChange based on request query params

QMalcolm:


soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
```diff
@@ -3,6 +3,8 @@ package com.socrata.soda.clients.datacoordinator
 import com.rojoma.json.v3.util.JsonUtil
 import com.rojoma.json.v3.ast._
 import scala.collection.Map
+import com.socrata.http.server.HttpRequest
+import scala.util.Try
```
rjmac: Never use Try

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
```diff
@@ -16,17 +18,32 @@ case class DeleteRow(rowId: JValue) extends RowUpdate {
...
+case class RowUpdateOptionChange(var truncate: Boolean = false,
+                                 var mergeInsteadOfReplace: Boolean = true,
+                                 var errorsAreFatal: Boolean = true,
+                                 var nonFatalRowErrors: Seq[String] = Seq())
+  extends RowUpdate {
+    def updateFromReq(req: HttpRequest) {
+      val parameterMap = req.servletRequest.getParameterMap()
+      parameterMap.get("truncate") match {
+        case null =>
+        case trunc => this.truncate = Try(trunc.asInstanceOf[Array[String]].head.toBoolean).getOrElse(false)
```
rjmac: _especially_ never use Try and then just throw away the failure case

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
```diff
@@ -16,17 +18,32 @@ case class DeleteRow(rowId: JValue) extends RowUpdate {
...
-//    if (!mergeInsteadOfReplace) map.put("update", JString("replace"))
-//    if (!errorsAreFatal) map.put("fatal_row_errors", JString(errorsAreFatal.toString))
-//    JObject(map)
-//  }
+// NOTE: If this class is changed, you may want to consider making similar changes
+// to RowUpdateOption.java of the soda-java project
+case class RowUpdateOptionChange(var truncate: Boolean = false,
+                                 var mergeInsteadOfReplace: Boolean = true,
+                                 var errorsAreFatal: Boolean = true,
+                                 var nonFatalRowErrors: Seq[String] = Seq())
```
rjmac: `var`s in case classes are at best extremely unidiomatic

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
QMalcolm: What should I be doing instead?

rjmac: ..also, don't use `.head`

rjmac: Also, don't use `asInstanceOf`.  You can do this directly in the match instead.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/RowDAO.scala
```diff
@@ -50,6 +50,9 @@ trait RowDAO {
   def upsert[T](user: String, datasetRecord: DatasetRecordLike, data: Iterator[RowUpdate], requestId: RequestId)
                (f: UpsertResult => T): T
 
+  def upsert[T](user: String, datasetRecord: DatasetRecordLike, data: Iterator[RowUpdate], requestId: RequestId, rowUpdateOption: RowUpdateOptionChange)
+              (f: UpsertResult => T): T
+
```
rjmac: It'd be better to not have this overload either always require `upsert` to take the parameter explicitly, or have it have a default value.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
```diff
@@ -35,6 +52,7 @@ case class RowUpdateOptionChange(truncate: Boolean = false,
       case true => JString("merge")
       case false => JString("replace")
     }),
-    "fatal_row_errors" -> JBoolean(errorsAreFatal)
+    "fatal_row_errors" -> JBoolean(errorsAreFatal),
+    "nonfatal_row_errors" -> JArray(nonFatalRowErrors.map(JString(_)))
```
rjmac: One of these will always override the other; I forget which though.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
rjmac: Catching only those exceptions you expect to be thrown.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
```diff
@@ -16,17 +18,32 @@ case class DeleteRow(rowId: JValue) extends RowUpdate {
...
+      }
+      parameterMap.get("mergeInsteadOfReplace") match {
+        case null =>
+        case mior => this.mergeInsteadOfReplace = Try(mior.asInstanceOf[Array[String]].head.toBoolean).getOrElse(true)
+      }
+      parameterMap.get("errorsAreFatal") match {
+        case null =>
+        case eaf => this.errorsAreFatal = Try(eaf.asInstanceOf[Array[String]].head.toBoolean).getOrElse(true)
+      }
+      parameterMap.get("nonFatalRowErrors[]") match {
```
rjmac: This convention has always felt really weird to me, but I don't really strongly object.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
```diff
@@ -16,17 +18,32 @@ case class DeleteRow(rowId: JValue) extends RowUpdate {
...
-//  }
+// NOTE: If this class is changed, you may want to consider making similar changes
+// to RowUpdateOption.java of the soda-java project
+case class RowUpdateOptionChange(var truncate: Boolean = false,
+                                 var mergeInsteadOfReplace: Boolean = true,
+                                 var errorsAreFatal: Boolean = true,
+                                 var nonFatalRowErrors: Seq[String] = Seq())
+  extends RowUpdate {
+    def updateFromReq(req: HttpRequest) {
+      val parameterMap = req.servletRequest.getParameterMap()
```
rjmac: I kind of feel like it'd be better to use the `parseQueryParemeterAs` function on `HttpRequest` instead of dropping down to the raw servlet level.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/RowUpdate.scala
```diff
@@ -16,25 +18,76 @@ case class DeleteRow(rowId: JValue) extends RowUpdate {
...
+          case true => NonFatalRowErrors(Nil)
+          case false => NoRowErrorsAreFatal
+        }
+      }
+    }
+
+    for {
+      truncate <- boolParam("truncate", false).right
+      mergeInsteadOfReplace <- boolParam("mergeInsteadOfReplace", true).right
+      errorPolicy <- errorPolicyParam().right
```
rjmac: Something's been sitting wrong with me here.  Since `default` was introduced, it'd probably be good to have `default.truncate` and `default.mergeInsteadOfReplace` in place of the `false` and `true` here.  The `errorPolicy` one can be deduped too but requires _not_ using `boolParam` in `errorPolicyParam` so it can handle the not-present case explicitly instead of deferring it.

rjmac: (in fact I'll just do that)

QMalcolm: Thanks :slightly_smiling_face:

PR #286, opened 2018-12-20
EN-29159 Need to remove location_state, city, etc in cjson row output.

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/LocationFuser.scala
```diff
@@ -7,13 +7,12 @@ import com.socrata.soda.server.highlevel.ExportDAO
 import com.socrata.soda.server.util.AdditionalJsonCodecs._
 import com.socrata.soda.server.wiremodels.{JsonColumnRep, JsonColumnWriteRep}
 import com.socrata.soql.types._
-
 import com.rojoma.json.util.JsonUtil
 
 /**
-  * Ouch, this class mutates the reps
+  * TODO: Not sure if this will work right for multiple locations.
```
bahrieinn: Confirmed this does not work yet. But we can address later

PR #288, opened 2019-03-13
Updated soda fountain to no longer drop the if-none-match header if i…

PiSoup:
…t is considered no longer valid by SF.


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
```diff
@@ -179,13 +179,23 @@ case class Resource(rowDAO: RowDAO,
         def prepareTag(etag: EntityTag) = etagObfuscator.obfuscate(etag.append(suffix))
         precondition.filter(_.endsWith(suffix)) match {
           case Right(newPrecondition) =>
+            // If there were originally etags in the precondition we need to add an invalid one back in
+            // so that lower services are aware that the request was made with the If-None-Match header
+            val updatedPrecondition = {
```
chitang: Let's make another function to keep the original function short
val updatedPrecondition = maybeUseFakeTag(newPrecondition)

define IfNoneOf(Seq(WeakEntityTag(Array(0)))) in Resource Object once

chitang:
:shipit:

PR #292, opened 2019-04-09
The datasets table has unused columns, ignore them

rjmac:
The primary key and schema hash columns are written to but henceforth
almost ignored, so make them ignorable.  Future update once this is
out will actually drop them.

The resolveSchemaInconsistency method actually used them, but it was
incorrect to do so.  I've changed it so it operates on a particular
copy, which is what it should have been doing all along.


chitang:
:shipit:

PR #294, opened 2019-04-25
EN-32009 Allow jnumber for text type in nbe

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/JsonColumnRep.scala
```diff
@@ -384,7 +399,7 @@ object JsonColumnRep {
 
   val forClientType: Map[SoQLType, JsonColumnRep] =
```
bahrieinn: can we add a test to JsonColumnRepTest.scala that shows the JNumber -> SoqlText thing if it doesn't already exist?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/JsonColumnRep.scala
```diff
@@ -209,6 +209,21 @@ object JsonColumnRep {
     val representedType: SoQLType = SoQLTime
   }
 
+  object ClientTextRep extends JsonColumnRep {
+    def fromJValue(input: JValue): Option[SoQLValue] = input match {
+      case JString(s) => Some(SoQLText(s))
+      case n: JNumber => Some(SoQLText(n.toString))
```
rjmac: ?!?! -- why?? This can only be a lurking trap for the unwary.  It's a terrible idea, not least because there is no unique string representation for any given number!

chitang: dozens of datasets from a dozens of customers datasets are affected after migration to nbe.  This will ease the path to nbe.

PR #302, opened 2019-06-06
EN-30766 allow writes to be gated on data version

rjmac:
Right now the source of this gate is a header on the request (just
like the output path, X-SODA2-Truth-Version) or, for upserts, a query
param expectedDataVersion because core uses soda-java as a
soda-fountain client when upserting and it's not obvious how to plumb
a header through soda-java without impacting the public API presented
there.


rjmac:
retest this please

PR #305, opened 2019-07-15
EN-33112 Encode resource name header for DC

rozap:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/ColumnDAOImpl.scala
```diff
@@ -323,14 +320,13 @@ class ColumnDAOImpl(dc: DataCoordinatorClient,
...
               } else {
-                val extraHeaders = Map(ReqIdHeader -> requestId,
-                                       SodaUtils.ResourceHeader -> dataset.name)
+                val extraHeaders = SodaUtils.traceHeaders(requestId, dataset)
                 dc.update(datasetRecord.systemId,
                           datasetRecord.schemaHash,
                           expectedDataVersion,
                           user,
                           Iterator.single(DropColumnInstruction(columnRef.id)),
-                          extraHeaders) {
```
QMalcolm: Why aren't we using the extraHeaders variable here?

rozap: good catch, was copy-pasting

PR #309, opened 2019-11-19
EN-36489 Fire rows_loaded_api metrics messages

chitang:
Not yet turned on.  Config will be done in docker files.

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/CJsonExporter.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/CJsonExporter.scala
@@ -1,6 +1,6 @@
 package com.socrata.soda.server.export
 
-import com.rojoma.json.v3.ast.{JString, JNumber}
+import com.rojoma.json.v3.ast.{JNumber, JString}
 import com.rojoma.json.v3.codec._
 import com.rojoma.json.v3.conversions._
 import com.rojoma.json.v3.io.CompactJsonWriter
@@ -13,9 +13,12 @@ import com.socrata.soda.server.util.AdditionalJsonCodecs._
 import com.socrata.soda.server.wiremodels.{JsonColumnRep, JsonColumnWriteRep}
 import com.socrata.soql.types.SoQLValue
 import java.io.BufferedWriter
+
 import javax.activation.MimeType
 import com.socrata.http.server.responses._
 import com.socrata.http.server.implicits._
+import com.socrata.soda.message.{MessageProducer, RowsLoadedApiMetricMessage}
+import com.socrata.soda.server.id.ResourceName
 
 object CJsonExporter extends Exporter {
   val mimeTypeBase = "application/json+x-socrata-cjson"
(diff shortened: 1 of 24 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/CJsonExporter.scala
```diff
@@ -81,7 +85,9 @@ object CJsonExporter extends Exporter {
         }
         w.write("]\n }\n")
         // end of header
+        var ttl = 0
```
PiSoup: Might want a less ambiguous variable name since "ttl" has meaning itself.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Export.scala
```diff
@@ -2,25 +2,26 @@ package com.socrata.soda.server.resources
...
 import com.socrata.http.server.routing.OptionallyTypedPathComponent
-import com.socrata.http.server.util.{Precondition, EntityTag, RequestId}
+import com.socrata.http.server.util.{EntityTag, Precondition, RequestId}
 import com.socrata.soda.server.copy.Stage
 import com.socrata.soda.server.responses._
 import com.socrata.soda.server.export.Exporter
-import com.socrata.soda.server.highlevel.ExportDAO.{ColumnInfo, CSchema}
-import com.socrata.soda.server.highlevel.{ExportParam, ColumnSpecUtils, ExportDAO}
+import com.socrata.soda.server.highlevel.ExportDAO.{CSchema, ColumnInfo}
+import com.socrata.soda.server.highlevel.{ColumnSpecUtils, ExportDAO, ExportParam}
```
PiSoup: A lot of noop import reorderings in here that you might want to revert.

chitang: i have a plugin that automatically re-order these.  it is probably fine to keep because what i introduced new isn't terribly complicated.  i don't strongly feel to the need to minimize unnecessary change.

chitang:
I copied from Alexa's work in DC or secondary watcher.  I tried to pull that into a library so that it can be shared among projects and I sort of did.  But the one in SW is targeting Eurybates and I was asked not to add raw messages capability in Eurybates lib.  The code is very similar but I can't factorize for both SW and SF.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/export/GeoJsonExporter.scala
```diff
@@ -31,13 +34,13 @@ object GeoJsonExporter extends Exporter {
...
              bom: Boolean = false,
-             fuseMap: Map[String, String] = Map.empty): HttpResponse = {
+             fuseMap: Map[String, String] = Map.empty)
+            (messageProducer: MessageProducer, entityIds: Seq[String], accessType: Option[String]): HttpResponse = {
     val mt = new MimeType(mimeTypeBase)
     mt.setParameter("charset", charset.alias)
     exporterHeaders(schema) ~> Write(mt) { rawWriter =>
       using(new BufferedWriter(rawWriter, 65536)) { w =>
-        val processor = new GeoJsonProcessor(w, schema, singleRow, obfuscateId)
-        processor.go(rows)
```
PiSoup: Did you intend to remove this?

PR #311, opened 2019-11-27
EN-35446: Limits the maximum number of threads used by data coordinat…

PiSoup:
…or to be 90% of the pool

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/CuratedHttpDataCoordinatorClient.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/CuratedHttpDataCoordinatorClient.scala
@@ -13,9 +13,15 @@ class CuratedHttpDataCoordinatorClient(httpClient: HttpClient,
                                        discoveredInstances: () => Set[String],
                                        serviceName: String,
                                        connectTimeout: FiniteDuration,
-                                       receiveTimeout: FiniteDuration)
+                                       receiveTimeout: FiniteDuration,
+                                       maxJettyThreadPoolSize: Int,
+                                       maxThreadRatio: Double)
   extends HttpDataCoordinatorClient(httpClient) with Closeable
 {
+  // Make sure the DC connection doesn't use all available threads
+  override val maxThreads = Some((maxThreadRatio * maxJettyThreadPoolSize).toInt)
+  override val consumerName = "DataCoordinatorClient"
+
   private[this] val connectTimeoutMS = connectTimeout.toMillis.toInt
   if (connectTimeoutMS != connectTimeout.toMillis) {
     throw new IllegalArgumentException("Connect timeout out of range (milliseconds must fit in an int)")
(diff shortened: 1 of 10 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/HttpDataCoordinatorClient.scala
```diff
@@ -463,6 +466,7 @@ abstract class HttpDataCoordinatorClient(httpClient: HttpClient) extends DataCoo
       val request = secondaryUrl(host, secondaryId, datasetId)
         .addHeader(("Content-type", "application/json"))
         .addHeaders(extraHeaders)
+        .timeoutMS(10000)
```
PiSoup: Unrelated to the thread problem, just missed it on a previous bugfix.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/CuratedHttpDataCoordinatorClient.scala
```diff
@@ -13,9 +13,13 @@ class CuratedHttpDataCoordinatorClient(httpClient: HttpClient,
                                        discoveredInstances: () => Set[String],
                                        serviceName: String,
                                        connectTimeout: FiniteDuration,
-                                       receiveTimeout: FiniteDuration)
+                                       receiveTimeout: FiniteDuration,
+                                       threadPool: Int)
   extends HttpDataCoordinatorClient(httpClient) with Closeable
 {
+  // Only allow data coordinator to use 90% of the available threads
+  override val maxThreads = Some((0.9 * threadPool).toInt)
```
znep: I'd probably be more conservative/aggressive here and set each of them to somewhere between 1/2 and 2/3rds.  If you are worried that will be too disruptive then I'd bump up the (overly conservative) max-threads from 100 to something higher to account for that.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/querycoordinator/CuratedHttpQueryCoordinatorClient.scala
```diff
@@ -11,9 +11,13 @@ class CuratedHttpQueryCoordinatorClient(val httpClient: HttpClient,
                                         discovery: ServiceDiscovery[AuxiliaryData],
                                         serviceName: String,
                                         connectTimeout: FiniteDuration,
-                                        receiveTimeout: FiniteDuration)
+                                        receiveTimeout: FiniteDuration,
+                                        threadPool: Int)
```
znep: nit: I'd call this something more explicit like `maxJettyThreadPoolSize` or some such.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/ThreadLimiter.scala
```diff
@@ -0,0 +1,35 @@
...
+        log.error(s"Trying to use threadpool without setting a maximum thread count")
+        f
+    }
+  }
+
+  def limitThreads[T](maxThreads: Int, threadCounter: AtomicInteger, f: => T): T = {
+    var response: Option[T] = None
+    try {
+      if (threadCounter.incrementAndGet() > maxThreads)
+        throw new Exception("Too many threads used")
```
znep: Is the resulting error message that gets logged relatively easy to trace to if it was dc or sf and what request it was for?  

Bonus points if you could make it return a reasonable HTTP error code (429?).

znep:
I'd probably also make the threshold a config value.

PR #312, opened 2019-11-29
Enable HTTP content compression on requests to qc / dc.

znep:
The CPU overhead is fairly minimal and the network cost adds up quickly.


chitang:
:shipit:

PR #313, opened 2019-12-13
Update SBT and Scala

rjmac:


soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorErrors.scala
```diff
@@ -274,16 +274,3 @@ object PossiblyUnknownDataCoordinatorError {
...
-
-
-
-
-
-
-
-
-
-
```
QMalcolm: This is my favorite part

PR #320, opened 2020-02-13
EN-38315: Don't try to snapshot deleted datasets.

znep:
Currently snapshotter is looping constantly trying to snapshot datasets
that are deleted, and failing because they can't be accessed because they
are marked deleted.  To avoid this we filter out deleted datasets.

The only other uses of bulkDatasetLookup are in collation related code, and
they don't make sense for deleted datasets either and/or already won't work
if the dataset is deleted.  I'm adding it as a paramter instead of
hardcoding it to help any future users of the method think about if they
want to include deleted datasets or not.

### DIFF
--- a/soda-fountain-lib/src/main/resources/com/socrata/soda/server/persistence/pg/sql/computation_strategies_create.sql
+++ b/soda-fountain-lib/src/main/resources/com/socrata/soda/server/persistence/pg/sql/computation_strategies_create.sql
@@ -5,6 +5,8 @@ CREATE TABLE IF NOT EXISTS computation_strategies (
   recompute                             BOOLEAN NOT NULL,
   source_columns                        TEXT[],
   parameters                            TEXT,
-  UNIQUE (dataset_system_id, column_id),
-  FOREIGN KEY (dataset_system_id, column_id) REFERENCES columns (dataset_system_id, column_id)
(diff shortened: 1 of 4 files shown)


soda-fountain-lib/src/main/resources/com/socrata/soda/server/persistence/pg/sql/computation_strategies_create.sql
```diff
@@ -5,6 +5,8 @@ CREATE TABLE IF NOT EXISTS computation_strategies (
   recompute                             BOOLEAN NOT NULL,
   source_columns                        TEXT[],
   parameters                            TEXT,
-  UNIQUE (dataset_system_id, column_id),
-  FOREIGN KEY (dataset_system_id, column_id) REFERENCES columns (dataset_system_id, column_id)
-);
\ No newline at end of file
+  UNIQUE (dataset_system_id, column_id)
```
rozap: changing migrations seems scary, what's the reasoning behind changing an old one vs creating a new one wiht the `alter table` statement?

znep:
I’d need to somehow find out the name of the constraint to be able to modify it, and the entire problem is the default name changes between postgres versions which is why the existing subsequent migration to change it fails.

In addition, the new migration would have to be inserted in the middle after this one but before subsequent ones to avoid them failing.

PR #328, opened 2020-07-21
EN-41094: pass relevant headers downstream always

rjmac:
This is a complex one.  This is making it so that request IDs are _always_ passed down to the lower levels of the stack (DC, QC) without needing to remember to do it.  It's done by making the http client inject the request id automatically.  This has some knock-on effects:

* The coordinator clients need to use the per-request http client, so they get instantiated per-request.  The curator management setup is still done eagerly, though.  The per-request part is done lazily, so if there's no need for the client, it's not created.
* The various DAOs need to use the per-request clients, so they also get instantiated per-request.  This too is done lazily.
* Similarly to the request IDs, the data coordinator client wants to send the resource name in a header for logging purposes, in addtion to the request ID.  This is done by making a `DatasetHandle` type that encapsulates both and making it take that.  Most things don't actually care they much, they just pass the id out of a `DatasetRecord` - since that has both things the DC client wants, it can now produce a `DatasetHandle` so most things just change `ds.datasetId` to `ds.handle` as the thing they pass to the client.

There are also some cleanups that are less related to this main project.  The big ones are
* making the {QC, DC} clients not inherit from `ThreadLimiter`.  Since `ThreadLimiter` is specifically _not_ per-request, it is now broken out into its own thing, and the clients share one (well two, one for each client type) among themselves.
* lots of little things to make the use of socrata-http-client a bit nicer.  More things now take an `HttpRequest` instead of an `HttpServletRequest` and as a result there's less dealing with nulls (since `HttpRequest` does all that for you)


rjmac:
This finished - https://jenkins-build.socrata.com/job/data-platform/job/soda-fountain/job/Soda%20Fountain%20Build%20Pipeline/job/PR-328/1/console - I don't know why Jenkins failed to tell GH about it.

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorClient.scala
```diff
@@ -132,73 +132,63 @@ object DataCoordinatorClient {
 trait DataCoordinatorClient {
   import DataCoordinatorClient._
 
-  def propagateToSecondary(datasetId: DatasetId,
-                           secondaryId: SecondaryId,
-                           extraHeaders: Map[String, String] = Map.empty)
-  def getSchema(datasetId: DatasetId): Option[SchemaSpec]
+  def propagateToSecondary(datasetId: DatasetHandle,
```
znep, line 135: nit: Should all these `datasetId` variable names also be updated to be either just `dataset` or `datasetHandle`?

soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
```diff
@@ -245,7 +242,7 @@ class DatasetDAOImpl(dc: DataCoordinatorClient,
   def getSecondaryVersions(dataset: ResourceName, requestId: RequestId): Result =
```
znep, line 242: Should all these `requestId` params on various methods go away too on all the `DAO` and `DAOImpl`s?

soda-fountain-lib/src/main/scala/com/socrata/soda/clients/datacoordinator/DataCoordinatorClient.scala
rjmac: Probably yeah.  I'll do a quick find/replace on them...

soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
rjmac: Almost certainly yes.  I thought I got them all but  I either missed quite a few or accidentally dropped it off my mental to-do list.

PR #333, opened 2020-08-20
EN-41655: Do some alias analysis when getting rollups too

rjmac:


rjmac:
It already does this analysis when creating rollups; it needs to do it when getting rollups to (the rollups are stored with column IDs, the creation phase converts column names to ids in an alias-aware way, so the get needs to do the opposite)

PR #341, opened 2021-04-13
EN-45573 SoQLJson 

rozap:
initial support for the json data type
* to be merged after https://github.com/socrata-platform/soql-postgres-adapter/pull/393

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/CsvColumnRep.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/CsvColumnRep.scala
@@ -212,6 +212,14 @@ object CsvColumnRep {
     }
   }
 
+  object JsonRep extends CsvColumnRep {
+    def toString(value: SoQLValue) =
+      value match {
+        case SoQLNull => null
+        case SoQLJson(jv) => JsonUtil.renderJson(jv)
+        case _ => null
+      }
+  }
 
   /**
     * Core is not supposed to use csv serialization format of photo type directly.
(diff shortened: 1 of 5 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/JsonColumnRep.scala
```diff
@@ -153,6 +153,23 @@ object JsonColumnRep {
...
+    }
+
+    def toJValue(input: SoQLValue): JValue = {
+      input match {
+        case SoQLJson(jv) => jv
+        case SoQLNull => JNull
+        case _ => stdBadValue
+      }
+    }
+  }
```
rjmac, line 170: This should do the (un)wrap-in `{json:...}` thing, I think.  Since this is just the middleman it won't matter too much, but it feels more correct than just being a passthrough.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/wiremodels/CsvColumnRep.scala
```diff
@@ -212,6 +212,14 @@ object CsvColumnRep {
     }
   }
 
+  object JsonRep extends CsvColumnRep {
+    def toString(value: SoQLValue) =
+      value match {
+        case SoQLNull => null
+        case SoQLJson(jv) => JsonUtil.renderJson(jv)
```
rjmac, line 219: Possibly `pretty=false` on that `renderJson`?

rjmac (approved):

PR #342, opened 2021-05-14
EN-46154: Do not produce scientific notation on SoQLNumber

chitang:
0.0000005302 should be written as is (not 5.302E-7)


saraid (approved):
String is so plaaaaaain

PR #343, opened 2021-06-10
EN-6239: Make computation strategy validation error 400 instead of 500

chitang:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/ColumnDAOImpl.scala
```diff
@@ -95,6 +96,10 @@ class ColumnDAOImpl(dc: DataCoordinatorClient,
             ColumnDAO.PreconditionFailed(f)
         }
       // TODO other cases have not been implemented
+      case InvalidComputationStrategy(err) =>
+        ColumnDAO.ComputationStrategyValidationError(err)
+      case err@WrongDatatypeForComputationStrategy(_, _) =>
+        ColumnDAO.ComputationStrategyValidationErrorResult(err)
       case err@x =>
         log.warn(s"case is NOT implemented ${err.getClass.getName}")
         ???
```
vickingan, line 105: I didn't know `???` this was legal in scala 🙃

chitang: it stands for missing implementation

PR #344, opened 2021-06-11
EN-6239: Make some validation error 400 instead of 50…

chitang:
…0, more

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/DatasetColumn.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/DatasetColumn.scala
@@ -68,6 +68,8 @@ case class DatasetColumn(etagObfuscator: ETagObfuscator, maxDatumSize: Int) {
         BadRequest ~> Json(error.message)
       case ColumnDAO.ComputationStrategyValidationErrorResult(result) =>
         BadRequest ~> Json(result.getClass.getSimpleName)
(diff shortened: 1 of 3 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/DatasetColumn.scala
```diff
@@ -68,6 +68,8 @@ case class DatasetColumn(etagObfuscator: ETagObfuscator, maxDatumSize: Int) {
         BadRequest ~> Json(error.message)
       case ColumnDAO.ComputationStrategyValidationErrorResult(result) =>
         BadRequest ~> Json(result.getClass.getSimpleName)
+      case ColumnDAO.ColumnValidationError(result) =>
+        BadRequest ~> Json(result.getClass.getSimpleName)
```
catstavi, line 72: is this an NBE response that core will need to be taught to handle to complete the transformation into a 400 response?

chitang: No change in core is needed.  It will be relayed upstream.

catstavi:
Just curious, this looks like its generally about invalid fieldnames, not computation strategy? or am I missing something?

chitang:
Title is miss leading.  I think this isn't necessarily computational strategy related.

catstavi (approved):

PR #358, opened 2022-02-15
EN-45452: Remove access type header, its usage, and Balboa metrics producing infrastructure

housepage:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
```diff
@@ -49,7 +44,6 @@ case class Resource(etagObfuscator: ETagObfuscator,
 
   val domainIdHeader = "X-SODA2-Domain-Id"
   val lensUidHeader = "X-Socrata-Lens-Uid"
-  val accessTypeHeader = "X-Socrata-Access-Type"
```
housepage, line 52: It looked like we could also probably phase out the `X-Socrata-Lens-Uid` header but I was less sure about that one.

rjmac (approved):

PR #364, opened 2022-03-03
EN-51020: Missing invalid rollups

chitang:


chitang:
This is just plumbing to get the new InvalidRollup error from DC back up the stack.

catstavi (approved):

PR #365, opened 2022-03-11
Give soda-fountain-external codecs

zachlefevre:
The purpose of this PR is to move some valuable SoqlType codecs (json and csv) to soda-fountain-external. When soda-fountain-external is published the archival-secondary will be able to access those codecs.


zachlefevre:
I can change the package names for the JsonColumnRep and CsvColumnRep. I just want to know if this is a reasonable approach.

soda-fountain-lib/src/test/scala/com/socrata/soda/server/wiremodels/JsonColumnRepPerfBench.scala
```diff
@@ -1,61 +0,0 @@
-package com.socrata.soda.server.wiremodels
```
zachlefevre, line 1: This is unused and doesn't contain any executable tests.

rozap (approved):

rozap:
I guess I wouldn't worry too much about changing names of things. 

I think our only goal here is to reduce the chance of breaking the internal-to-socrata load path that users can take. Ie: the dsmapi view source path and the dsmapi archival source path. So the concerns around this code being something that shouldn't exist, while valid, aren't really something that is in scope here. Which is to say, names and places for functionality I think will always be a little weird, and I'm not sure there's a "right" way outside of going back in time and finished projects better.

zachlefevre:
@catstavi I'm not sure. I moved only those codecs which we want to share externally (specifically with the archival secondary) an left everything else in place. I'd prefer to only move those modules which are needed externally into the `external` package, but I understand it's strange that some of the `reprs` are in the external package and some are not.

Do you have a preference?

catstavi (approved):
sounds reasonable for the moment, and like you've been seeking feedback from sources more in touch with this service

PR #366, opened 2022-03-14
Make Jenkins publish external library

zachlefevre:


catstavi (commented):
normally I think I would expect something that triggers downstream job 'sbt build', but I guess soda-fountain doesn't even use the shared dockerize job? I don't see any problems with this, but I'm also in no way an expert on this jenkins file or soda fountain build procedures. 

Have you tested building this branch in jenkins?

zachlefevre:
@catstavi It does use the shared dockerize job https://github.com/socrata-platform/soda-fountain/blob/zlef%2Fmake_jenkins_publish_external_lib/Jenkinsfile#L169

zachlefevre:
@catstavi and this will kick off the shared sbt build job https://github.com/socrata-platform/soda-fountain/blob/zlef%2Fmake_jenkins_publish_external_lib/Jenkinsfile#L21

zachlefevre:
I haven't tested it directly since doing so would require cutting a release of SF.

catstavi (approved):
none of the jobs we're trying run take a branch specifier (for easy testing) and also its a pipeline build structure I haven't seen before! so I have no step by steps for testing this job... we can always revert I suppose

PR #370, opened 2022-04-07
Resync API Endpoint

zachlefevre:


rozap (approved):
looks good. have you already done the data-coordinator side of things?

rozap:
nope, i think they can be merged in either order. just that i'd look at them together if it was there, since i think the integration is probably the more fiddly part.

PR #372, opened 2022-04-15
EN-52670: Add field rollup raw_soql

chitang:


saraid (approved):
I don't know enough about this to have a problem with it.

PR #373, opened 2022-04-15
Defense against invalid rollup soql

chitang:


rjmac (approved):
Removing dead code and catching BadParse?  LGTM! :shipit:

PR #379, opened 2022-06-15
EN-54122: update jenkinsfile for releasing with no changes

sarah-salinas:
## Before

For release cuts, it would exit if there were no changes and the "FORCE_BUILD" option was false.
If the "FORCE_BUILD" option was true, then it would fail in the "Publish" stage because artifactory doesn't let us publish if there are no changes.
However, we now need to rebuild and deploy even if there are no changes to the repo to pick up base layer changes for aqua compliance.

## After
For release cuts, if there are no changes, it will rebuild from the current tag and skip the publish stage.
I removed the "FORCE_BUILD" logic since we want that to happen every time "RELEASE_CUT" is true.  I'll remove the "FORCE_BUILD" parameter from the cut job once this is merged.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,7 +28,6 @@ pipeline {
   }
   parameters {
     booleanParam(name: 'RELEASE_CUT', defaultValue: false, description: 'Are we cutting a new release candidate?')
-    booleanParam(name: 'FORCE_BUILD', defaultValue: false, description: 'Force build from latest tag if sbt release needed to be run between cuts')
     string(name: 'AGENT', defaultValue: 'build-worker-pg13', description: 'Which build agent to use?')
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -59,9 +58,8 @@ pipeline {
           service_sha = sh(returnStdout: true, script: "git rev-parse HEAD").trim()
 
           // determine what triggered the build and what stages need to be run
-          if (params.RELEASE_CUT == true) { // RELEASE_CUT parameter was set by a cut job
+          if (params.RELEASE_CUT) { // RELEASE_CUT parameter was set by a cut job
             stage_cut = true  // other stages will be turned on in the cut step as needed
```
zachlefevre, line 62: we unconditionally turn on those stages in the stage_cut phase. Should we pull those up here for consistency?

sarah-salinas: Good idea.

PR #382, opened 2022-08-16
Expand error (type) handling when bad rollup is encountered

chitang:
Previously we only catch parse error.  But there is more work after parsing that has potential to err.


saraid (approved):
Warning is good, yeah.

PR #383, opened 2022-09-23
EN-54831 - pass lens id through to query coordinator

branweb1:
Query coordinator needs the lens id for soql analysis. Core sends it but soda fountain was swallowing it.

### DIFF
--- a/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
+++ b/soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
@@ -207,7 +207,8 @@ case class Resource(etagObfuscator: ETagObfuscator, maxRowSize: Long, metricProv
                     req.header("X-Socrata-Explain").isDefined,
                     req.header("X-Socrata-Analyze").isDefined
                   ),
-                  req.resourceScope) match {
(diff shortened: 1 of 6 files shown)


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
```diff
@@ -207,7 +207,8 @@ case class Resource(etagObfuscator: ETagObfuscator, maxRowSize: Long, metricProv
                     req.header("X-Socrata-Explain").isDefined,
                     req.header("X-Socrata-Analyze").isDefined
                   ),
-                  req.resourceScope) match {
+                  req.resourceScope,
+                  req.header("X-Socrata-Lens-Uid").get) match {
```
PaulMTT, line 211: .get will throw an exception if this value is missing.
Is throwing an exception at this code point a valid action?

PaulMTT (approved):

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Resource.scala
branweb1: after thinking about it, I think no. Made this optional

PR #402, opened 2023-06-09
EN-60548: Fake-locations on the new query path

rjmac:
When told we're looking at a location column, find its subcolumns and pass this information down to query-coordinator


saraid (approved):
Not fond of `Aux`, really,  but I guess it works.

PR #423, opened 2024-02-12
New-analyzer rollups are not invalid

rjmac:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
```diff
@@ -500,41 +500,58 @@ class DatasetDAOImpl(dc: DataCoordinatorClient,
...
-                  RollupSpec(name = rollup.name, soql = "__Invalid SoQL__")
-                case ex: Exception =>
-                  log.warn(s"invalid rollup SoQL ${rollup.name} ${rollup.soql} ${ex.getMessage}")
-                  RollupSpec(name = rollup.name, soql = "__Invalid Rollup__")
+              if(rollup.isNewAnalyzer) {
+                // New-rollups don't get reworked, so we can just hand back the soql unmodified
+                RollupSpec(
+                  name = rollup.name,
+                  soql = rollup.soql
+                )
```
rjmac, line 508: Oh hrm actually this isn't quite true.  I'll have to think about this a bit more (it _is_ true that the soql per se doesn't get rewritten, but all the pointers do)

soda-fountain-lib/src/main/scala/com/socrata/soda/server/persistence/pg/PostgresStoreImpl.scala
```diff
@@ -27,6 +27,153 @@ import org.joda.time.DateTime
...
+       |       cs.parameters
+       | FROM columns c
+       | JOIN dataset_copies dc on dc.id = c.copy_id
+       | LEFT JOIN computation_strategies cs
+       | ON c.dataset_system_id = cs.dataset_system_id AND c.column_id = cs.column_id AND c.copy_id = cs.copy_id
+       | WHERE $columnFilter
+        |dc.dataset_system_id = ? AND
+        |       dc.copy_number = ?
+        |       $deletedFilter
+        |       """.stripMargin
```
saraid, line 49:
```suggestion
       | dc.dataset_system_id = ? AND
       |        dc.copy_number = ?
       |        $deletedFilter
       |        """.stripMargin
```
May as well fix it now.

saraid (approved):
Couldn't spot any problems, though I'm kinda irritated by all the weirdly-cased SQL that got shuffled around.

rjmac:
> Couldn't spot any problems, though I'm kinda irritated by all the weirdly-cased SQL that got shuffled around.

Did that because of implicit-resolution (all uses of an implicit in the same file where it's defined must come after the definition)

PR #424, opened 2024-03-07
EN-62410: RMS integration

sarah-salinas:
This commit adds integration for the release metadata service for release builds.

See https://github.com/socrata-platform/soql-postgres-adapter/pull/646

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,14 @@
 // Set up the libraries
 @Library('socrata-pipeline-library')
 
+import com.socrata.ReleaseMetadataService
+def rmsSupportedEnvironment = com.socrata.ReleaseMetadataService.SupportedEnvironment
+
 // set up variables
-def service = 'soda-fountain'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,11 +1,14 @@
...
+
 // set up variables
-def service = 'soda-fountain'
-def project_wd = 'soda-fountain-jetty'
-def isPr = env.CHANGE_ID != null
-def lastStage
+String service = 'soda-fountain'
+String project_wd = 'soda-fountain-jetty'
+boolean isPr = env.CHANGE_ID != null
+String lastStage
```
ds2643, line 11: You didn't introduce the `lastStage` definition, but I think departing from the variable-instantiation-without-a-binding pattern in Groovy would be beneficial in avoiding null pointer exceptions. I recommend refraining from defining variables without bindings. I'm not sure if the default binding is `""` or `null`.

ds2643 (approved):
These changes are largely the same as those reviewed elsewhere across this integration.

PR #433, opened 2024-05-03
EN-66368: Use resource names instead of internal names in new-rollups

rjmac:


rlvoyer:
Looks good to me, but a little description for the merge commit would be good for posterity.

rlvoyer (approved):
Looks good to me, but a little description for the merge commit would be good for posterity.

PR #441, opened 2024-08-20
integrating resource groups

dmitry-at-tyler:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Dataset.scala
```diff
@@ -182,6 +184,7 @@ case class Dataset(maxDatumSize: Int) {
 
     override def delete = { req =>
       val datetime = req.dateTimeHeader("X-Socrata-Deleted-At") // allow expedited delete
+      resourceGroupsClient.deregisterDataset(resourceName.toString())
```
rjmac, line 187: This method doesn't _actually_ delete the dataset; it just marks it for later deletion in order to support dataset restore.  This should happen when the dataset is deleted for real some time later.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resource_groups/ResourceGroupsClientFactory.scala
```diff
@@ -0,0 +1,24 @@
...
+}
+
+class ResourceGroupsClientFactory private(apiHost: String) {
+  private def _client(): ResourceGroupsClient =
+    ResourceGroupsClientBuilder.builder()
+      .apiHost(apiHost)
+      .httpClientAdapter(new ResourceGroupsHttpAdapter())
+      .jsonCodecAdapter(new ResourceGroupsJsonAdapter())
+      .build()
+}
```
rjmac, line 24: The way SF wants to work is for the RGCF to be built once in main and then passed into everything that needs it, rather than being a global.  That would also mean it could receive its configuration as a parameter rather than using a global - and only the configuration that it wants (i.e., the resource groups api host) rather than All The Config

rjmac (changes requested):
See above

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resource_groups/ResourceGroupsClientFactory.scala
rjmac: (it'd also mean it could use the common http client that's built there rather than its own,  much as core's RG http adapter does)

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resources/Dataset.scala
dmitry-at-tyler: yes; and the intention is that once it's marked for deletion, resource groups are not tracking it anymore.

soda-fountain-lib/src/main/scala/com/socrata/soda/server/resource_groups/ResourceGroupsClientFactory.scala
PaulMTT: Yes but plumbing this all down is not straight forward for our case where we targeted the dataset remove. I have reworked all existing places that relied on `new SodaFountainConfig()`, seemed the most direct for now. Open to suggestions though.

PaulMTT: Alright I have passed this through to datasetdaoimpl, where we are marking this deregistered there.

PaulMTT:
Deps are threaded through, including socrata http lib.
@rjmac Please take another look!

rjmac (approved):

PR #445, opened 2024-08-22
Paulm/resource groups config

PaulMTT:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/highlevel/DatasetDAOImpl.scala
```diff
@@ -226,6 +227,7 @@ class DatasetDAOImpl(dc: DataCoordinatorClient,
       store.translateResourceName(dataset) match {
         case Some(datasetRecord) =>
           store.markResourceForDeletion(dataset, datetime)
+          rg.deregisterDataset(dataset.toString())
```
dmitry-at-tyler, line 230: try/catch this ?

dmitry-at-tyler (approved):

PR #458, opened 2024-11-01
EN-70113: Plumb resource groups api protocol config

PaulMTT:


soda-fountain-lib/src/main/scala/com/socrata/soda/server/resource_groups/ResourceGroupsClientFactory.scala
```diff
@@ -1,15 +1,22 @@
...
 import com.socrata.http.client.HttpClient
-import com.socrata.resource_groups.client.{ResourceGroupsClient, ResourceGroupsClientBuilder}
+import com.socrata.resource_groups.client.{ResourceGroupsClient, ResourceGroupsClientBuilder, ResourceGroupsException}
 import com.socrata.soda.server.config.{ResourceGroupsClientConfig, SodaFountainConfig}
 
 
 class ResourceGroupsClientFactory(resourceGroupClientConfig: ResourceGroupsClientConfig, httpClient: HttpClient) {
   private lazy val _client = {
-    println(s"Resource groups apiHost: '${resourceGroupClientConfig.apiHost}'")
+    val apiHost = resourceGroupClientConfig.apiHost match {
```
dmitry-at-tyler, line 10: I think it would be better to make the change to the resource groups client, and add separate "protocol" configuration.

PaulMTT: We could definitely do that, but I like to think of urls as a singular thing protocol included. We are just doing this to work around an apps-marathon config issue. We are passing config into the factory as a config object group, so that's good.

dmitry-at-tyler (approved):

dmitry-at-tyler (approved):

PR #465, opened 2025-04-04
Report "schema out of sync" errors correctly

rjmac:


saraid (approved):
I wonder how this happened in the first place...

